PR #8, opened 2025-10-15
Runtime cache sync/async support

ricardo-agz:
Sync and Async runtime cache support now available via:
```python
from vercel.cache import get_cache
from vercel.cache.aio import get_cache as aget_cache
cache = get_cache()
acache = aget_cache()
cache.set(..)
await acache.get(...)
```

**using clients**
```python
from vercel.cache import RuntimeCache, AsyncRuntimeCache
cache = RuntimeCache()
acache = AsyncRuntimeCache()
cache.set(..)
await acache.get(...)
```


brookemosby (approved):
LGTM

PR #14, opened 2025-10-16
projects & deployments sync + async support

ricardo-agz:
**projects sync + async support**
```python
from vercel.projects import create_project
from vercel.projects.aio import create_project as acreate_project

create_project(...)
await acreate_project(...)
```

**with clients**
```python
from vercel.client import Vercel, AsyncVercel

vc_client = AsyncVercel()

vc_client.deployments.create_deployment(...)
```

Also adds convenience methods to sync runtime cache to enable
```python
if key in cache:
    print(cache[key])
```

### DIFF
--- a/src/vercel/__init__.py
+++ b/src/vercel/__init__.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+from .client import Vercel, AsyncVercel
+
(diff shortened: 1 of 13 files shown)


brookemosby (approved):
LGTM

PR #31, opened 2025-10-27
updating schemas

brookemosby:


src/vercel/telemetry/tracker.py
```diff
@@ -0,0 +1,130 @@
...
+    if _telemetry_client is None:
+        try:
+            from .client import TelemetryClient
+            _telemetry_client = TelemetryClient()
+        except Exception:
+            pass
+    return _telemetry_client
+
+
+def _track_safely(func_name: str, **kwargs: Any) -> None:
```
ricardo-agz: I think it's totally fine to just keep this function and expose it as a generic `track(...)` function that we can call from wherever we want to track stuff so we don't have to worry about maintaining parity in this file with every future package function we add or having some functions have a dedicated `track_*()` function but others using `track()`, etc. 

Something like:
```python
def get_client() -> TelemetryClient | None:
    global _telemetry_client
    if _telemetry_client is None:
        try:
            _telemetry_client = TelemetryClient()
        except Exception:
            return None
    return _telemetry_client

def track(event: str, **attrs: Any) -> None:
    client = get_client()
    if client is None:
        return
    try:
        client.track(event, **attrs)  # add a generic .track() in client
    except Exception:
        pass
```

src/vercel/blob/ops.py
```diff
@@ -187,15 +196,19 @@ def delete(
...
         "POST",
         options={"token": token} if token else {},
         headers={"content-type": "application/json"},
         body={"urls": urls},
     )
+    # Track telemetry
+    track_blob_delete(count=count)
 
 
 async def delete_async(
```
ricardo-agz, line 327: Also, I think it would be quite nice if we could have a decorator way of tracking functions to avoid having to maintain a dedicated tracking function for each function, something like:

```python
def telemetry(
    event: str,
    capture: Sequence[str] | None = None,
    derive: Mapping[str, Callable[[tuple, dict, Any], Any]] | None = None,
    when: Literal["before", "after"] = "after",
):
    """Decorator to emit telemetry around a function call.

    - ``capture``: list of parameter names to capture from args/kwargs.
      Names are resolved against the function signature so positional calls
      are handled correctly.
    - ``derive``: mapping of output field -> lambda(args, kwargs, result).
    - ``when``: emit "before" the call, "after" the call.
    """

    def decorator(func: Callable[..., Any]):
        is_coro = inspect.iscoroutinefunction(func)
        sig = inspect.signature(func)

        def _emit(ev: str, args: tuple, kwargs: dict, result: Any) -> None:
            try:
                attrs: dict[str, Any] = {}
                # Bind parameters for positional resolution
                try:
                    bound = sig.bind_partial(*args, **kwargs)
                    params: dict[str, Any] = dict(bound.arguments)  # name -> value
                except Exception:
                    params = {}

                # Capture selected params by name
                if capture:
                    for name in capture:
                        if name in kwargs:
                            attrs[name] = kwargs[name]
                        elif name in params:
                            attrs[name] = params[name]
                        # else: silently skip if not provided

                # Derived attributes
                if derive:
                    for field, getter in derive.items():
                        try:
                            attrs[field] = getter(args, kwargs, result)
                        except Exception:
                            # ignore individual derivation errors
                            pass

                track(ev, **attrs)
            except Exception:
                # Silently fail - don't impact user's operation
                pass

        @functools.wraps(func)
        async def async_wrapper(*args: Any, **kwargs: Any):
            if when == "before":
                _emit(event, args, kwargs, None)
            result = await func(*args, **kwargs)
            if when == "after":
                _emit(event, args, kwargs, result)
            return result

        @functools.wraps(func)
        def sync_wrapper(*args: Any, **kwargs: Any):
            if when == "before":
                _emit(event, args, kwargs, None)
            result = func(*args, **kwargs)
            if when == "after":
                _emit(event, args, kwargs, result)
            return result

        return async_wrapper if is_coro else sync_wrapper

    return decorator
```

that way we could track functions something like:

```python
def derive_delete_count(args, kwargs, rv):
    v = kwargs.get("url_or_path", args[0] if args else None)
    if v is None or isinstance(v, (str, bytes)):
        return 1
    return len(v) if hasattr(v, "__len__") else 1

@telemetry(
    event="blob_delete",
    derive={"count": derive_delete_count},
    when="after",
)
def delete(url_or_path: str | Iterable[str], *, token: str | None = None) -> None:
    ...
```

src/vercel/telemetry/client.py
```diff
@@ -0,0 +1,329 @@
...
+            "project_create": [],
+            "project_update": [],
+            "project_delete": [],
+            "deployment_create": [],
+        }
+        self._enabled = _TELEMETRY_ENABLED
+        # Register flush at exit so telemetry events are sent before program termination
+        atexit.register(self._flush_at_exit)
+
+    def track_blob_put(
```
ricardo-agz: Similar to the other notes, I think it might be better to have a single `track(...)` function or `@telemetry(...)` decorator or something so that we don't have to maintain parity in this file when we add new packages/functions. Then the client would only need to have a single `client.track(...)` function

src/vercel/_telemetry/client.py
```diff
@@ -0,0 +1,329 @@
...
+        except Exception as e:
+            if _TELEMETRY_DEBUG():
+                print(f"Telemetry error: {e}")
+
+    def reset(self) -> None:
+        """Clear accumulated events."""
+        for event_type in self._events:
+            self._events[event_type].clear()
+
+    def _flush_at_exit(self) -> None:
```
ricardo-agz, line 168: This will fail silently if you call `asyncio.run()` inside an existing event loop. If we're only flushing at exit, I think it's probably fine to just make flush sync and call `flush()` regularly, otherwise we'd have to check to see if there's a running event loop and use that or handle that edge case somehow.

repro:
```python
import os
import asyncio

# Disable actual telemetry network calls for this repro
os.environ["VERCEL_TELEMETRY_DISABLED"] = "1"

from vercel.telemetry.client import TelemetryClient


async def main() -> None:
    tc = TelemetryClient()
    # Ensure there is at least one event queued
    tc.track_blob_put(size_bytes=1)

    print("1) Calling _flush_at_exit() inside a running event loop (exceptions are suppressed)...")
    tc._flush_at_exit()
    print("   ...no exception propagated because _flush_at_exit() swallows it.\n")

    print("2) Reproducing the underlying error explicitly by calling asyncio.run(tc.flush())\n"
          "   from within the running event loop:")
    try:
        asyncio.run(tc.flush())
    except Exception as e:
        # Expected: RuntimeError("asyncio.run() cannot be called from a running event loop")
        print(f"   Caught: {type(e).__name__}: {e}")


if __name__ == "__main__":
    asyncio.run(main())
```

```
(.venv) ricardo@Ricardos-MacBook-Pro examples % /Users/ricardo/Documents/code/vercel-python-sdks/vercel-py/
.venv/bin/python /Users/ricardo/Documents/code/vercel-python-sdks/vercel-py/examples/telemetry_event_loop_r
epro.py
1) Calling _flush_at_exit() inside a running event loop (exceptions are suppressed)...
/Users/ricardo/Documents/code/vercel-python-sdks/vercel-py/src/vercel/telemetry/client.py:323: RuntimeWarning: coroutine 'TelemetryClient.flush' was never awaited
  pass
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
   ...no exception propagated because _flush_at_exit() swallows it.

2) Reproducing the underlying error explicitly by calling asyncio.run(tc.flush())
   from within the running event loop:
   Caught: RuntimeError: asyncio.run() cannot be called from a running event loop
/Users/ricardo/Documents/code/vercel-python-sdks/vercel-py/examples/telemetry_event_loop_repro.py:25: RuntimeWarning: coroutine 'TelemetryClient.flush' was never awaited
  print(f"   Caught: {type(e).__name__}: {e}")
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
(.venv) ricardo@Ricardos-MacBook-Pro examples % 
```

src/vercel/blob/ops.py
```diff
@@ -186,15 +255,22 @@ def delete(
...
+        count = 1
     request_api(
         "/delete",
         "POST",
         token=token,
         headers={"content-type": "application/json"},
         body={"urls": urls},
     )
+    # Track telemetry
+    try:
```
ricardo-agz: nit: `track()` already has a try/catch so we can just call `track(...)` without having to re-wrap it in try/catch

src/vercel/telemetry/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Telemetry functionality for Vercel Python SDK."""
+
+from .client import TelemetryClient
+
+__all__ = ["TelemetryClient"]
```
ricardo-agz: let's also re-export `track()` here so we can do `from vercel._telemetry import track` from the sandbox-py repo
also, we should probably rename the folder to `/_telemetry` since it's for internal use

src/vercel/deployments/deployments.py
```diff
@@ -267,4 +276,12 @@ async def create_deployment_async(
         raise RuntimeError(
             f"Failed to create deployment: {resp.status_code} {resp.reason_phrase} - {data}"
         )
+    # Track telemetry
+    try:
```
ricardo-agz: nit: since we already declare the type of body to be a dict in the function signature, we can add a `if not isinstance(body, dict) raise ValueError(...)` as a validation to the top of the function and then get rid of the try/catch here and just do `target=body.get("target")`

src/vercel/projects/projects.py
```diff
@@ -212,6 +214,11 @@ def create_project(
         raise RuntimeError(
             f"Failed to create project: {resp.status_code} {resp.reason_phrase} - {data}"
         )
+    # Track telemetry
+    try:
```
ricardo-agz: same here, since track already try/catches internally we can get rid of the try/catches here

src/vercel/telemetry/credentials.py
```diff
@@ -0,0 +1,79 @@
...
+    if not resolved_project_id:
+        resolved_project_id = os.getenv("VERCEL_PROJECT_ID")
+    if not resolved_team_id:
+        resolved_team_id = os.getenv("VERCEL_TEAM_ID")
+    
+    # Try to extract from OIDC token if available
+    if token:
+        try:
+            # Import lazily to avoid hard dependency in all environments
+            from ..oidc.token import decode_oidc_payload as _decode_oidc_payload  # type: ignore
```
ricardo-agz: nit: oidc is part of the vercel package, I think it's fine to just import it regularly at the top of the file `from vercel.oidc import decode_oidc_payload`

ricardo-agz: nevermind, saw the top comment, if it causes mypy errors its fine

src/vercel/deployments/deployments.py
```diff
@@ -116,6 +117,14 @@ def create_deployment(
...
```
ricardo-agz: nit: similar comment here, since body is supposed to be a dict, we should add top level validation then get rid of try/catch and just do `target=body.get("target")`

ricardo-agz (approved):

PR #42, opened 2026-02-02
Add _core module with shared sync/async client using iter_coroutine

scotttrinh:
This is not the final form (we want pydantic models, etc), but this is the POC for using the `iter_coroutine` approach we discussed. How does this feel?

### DIFF
--- a/src/vercel/_core/client.py
+++ b/src/vercel/_core/client.py
@@ -1,0 +1,77 @@
+"""Vercel API clients with namespaced sub-clients."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from .config import DEFAULT_API_BASE_URL, DEFAULT_TIMEOUT, ClientConfig
+from .deployments import AsyncDeploymentsClient, DeploymentsClient
+from .iter_coroutine import iter_coroutine
+from .projects import AsyncProjectsClient, ProjectsClient
+from .transport import AsyncTransport, BlockingTransport
+
+
+class VercelClient:
+    """Synchronous Vercel SDK client."""
+
+    def __init__(
+        self,
+        *,
+        access_token: str | None = None,
+        base_url: str | None = None,
+        timeout: float | None = None,
+        default_team_id: str | None = None,
+        default_slug: str | None = None,
+    ):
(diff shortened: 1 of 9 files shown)


src/vercel/_core/transport.py
```diff
@@ -0,0 +1,120 @@
...
+        path: str,
+        *,
+        params: dict[str, Any] | None = None,
+        json: Any | None = None,
+        content: bytes | None = None,
+        headers: dict[str, str] | None = None,
+    ) -> httpx.Response: ...
+
+    @abc.abstractmethod
+    async def close(self) -> None: ...
```
1st1, line 32:
```suggestion
    async def close(self) -> None:
        raise NotImplementedError
```

src/vercel/_core/transport.py
```diff
@@ -0,0 +1,120 @@
...
+    async def send(
+        self,
+        method: str,
+        path: str,
+        *,
+        params: dict[str, Any] | None = None,
+        json: Any | None = None,
+        content: bytes | None = None,
+        headers: dict[str, str] | None = None,
+    ) -> httpx.Response: ...
```
1st1, line 29:
```suggestion
    ) -> httpx.Response: ...
    raise NotImplementedError
```

src/vercel/_core/client.py
```diff
@@ -0,0 +1,77 @@
...
+        iter_coroutine(self._transport.close())
+
+    def __enter__(self) -> VercelClient:
+        return self
+
+    def __exit__(self, *args: Any) -> None:
+        self.close()
+
+
+class AsyncVercelClient:
```
1st1, line 47:
```suggestion
class AsyncIOVercelClient:
```

and we also need to add a TrioVercelClient

1st1 (approved):
Aside from a couple of nits this is exactly where I wanted to go with this. Please push forward. I'd add "trio" support + tests and once we are there, I'd start converting the code + increasing the API coverage with those three targets (asycio, trio, blocking io) in mind.

scotttrinh:
Closing in favor of a different starting point

PR #44, opened 2026-02-03
Reduce duplication between sync/async

scotttrinh:
This is a no-op code reorganization. The main improvement being removing the duplication of business-logic and code between async and sync implementations using a method pioneered in https://github.com/geldata/gel-python that allows the main logic to be written once and for sync and async code to share that base logic while having the correct runtime behavior.

Since this is a large change, here are some special areas of note:

- We had different timeouts defined for sync vs async in our HTTP requests. I unified them here, but if there is a reason they were different, we can disintegrate that.
- The non-SDK modules have a best-effort attempt at this code reuse pattern, but I think we probably want a more thorough refactoring of our HTTP layer to support some of the more advanced use-cases that `vercel.blob`, for instance, relies on (retries, etc).
- I (we) added a full regression test suite that we ran against `main` which passed, which feels like it gives us _some_ semblance of safety that this is a no-op refactor.

---

Here is Opus' rendition of the overall approach here:

```
  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                         SHARED CODE PATTERN                                 │
  └─────────────────────────────────────────────────────────────────────────────┘

                                _core.py
                      ┌──────────────────────┐
                      │   _BaseAPIClient     │
                      │   ──────────────     │
                      │   _transport: Base   │  ◄── Abstract transport
                      │                      │
                      │   async def _get()   │  ◄── All business logic is async
                      │   async def _put()   │
                      │   async def _list()  │
                      └──────────┬───────────┘
                                 │
                ┌────────────────┴────────────────┐
                │                                 │
                ▼                                 ▼
  ┌─────────────────────────┐       ┌─────────────────────────┐
  │    SyncAPIClient        │       │    AsyncAPIClient       │
  │    ─────────────        │       │    ──────────────       │
  │ _transport: Blocking    │       │ _transport: Async       │
  └─────────────────────────┘       └─────────────────────────┘


  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                         TRANSPORT LAYER                                     │
  └─────────────────────────────────────────────────────────────────────────────┘

                      ┌──────────────────────┐
                      │    BaseTransport     │
                      │    ─────────────     │
                      │  async def send()    │  ◄── Always async interface
                      └──────────┬───────────┘
                                 │
                ┌────────────────┴────────────────┐
                │                                 │
                ▼                                 ▼
  ┌─────────────────────────┐       ┌─────────────────────────┐
  │   BlockingTransport     │       │    AsyncTransport       │
  │   ─────────────────     │       │    ──────────────       │
  │   httpx.Client          │       │   httpx.AsyncClient     │
  │                         │       │                         │
  │   async def send():     │       │   async def send():     │
  │     # uses iter_coro    │       │     # native async      │
  │     # to run sync       │       │     await client.req()  │
  └─────────────────────────┘       └─────────────────────────┘


  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                         iter_coroutine                                      │
  └─────────────────────────────────────────────────────────────────────────────┘

      Sync function call:                    Async function call:
      ───────────────────                    ────────────────────

      result = get_projects(...)             result = await get_projects_async(...)
           │                                      │
           ▼                                      ▼
      ┌─────────────────┐                   ┌─────────────────┐
      │ SyncClient      │                   │ AsyncClient     │
      │                 │                   │                 │
      │ iter_coroutine( │                   │ await           │
      │   self._get()   │                   │   self._get()   │
      │ )               │                   │                 │
      └────────┬────────┘                   └────────┬────────┘
               │                                     │
               ▼                                     ▼
      ┌─────────────────┐                   ┌─────────────────┐
      │ iter_coroutine  │                   │ Native async    │
      │ ──────────────  │                   │                 │
      │ Steps through   │                   │ Event loop      │
      │ coroutine with  │                   │ handles awaits  │
      │ coro.send()     │                   │ normally        │
      │                 │                   │                 │
      │ BlockingTransport                   │ AsyncTransport  │
      │ does sync HTTP  │                   │ does async HTTP │
      └─────────────────┘                   └─────────────────┘


  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                         WHY THIS WORKS                                      │
  └─────────────────────────────────────────────────────────────────────────────┘

    1. Business logic written ONCE as async methods in _Base class

    2. Transport abstraction means the same async code can use either:
       • BlockingTransport (sync httpx.Client)
       • AsyncTransport (async httpx.AsyncClient)

    3. iter_coroutine "drives" the coroutine manually for sync path:
       • Calls coro.send(None) repeatedly
       • Coroutine yields control at await points
       • But BlockingTransport.send() completes synchronously
       • So the coroutine runs to completion without an event loop

    4. Async path just uses normal await - no special handling needed

### CODE REVIEW

**Comment by scotttrinh** [comment]
Looks like there are a few issues to fix here. I'm going back to draft and removing reviewers and will fix it up in the morning.

File: src/vercel/_http/transport.py [hunk 1]
```diff
@@ -63,23 +83,36 @@ async def send(
...
+            request_headers.update(headers)
+
+        # Unpack content based on type
+        json_data: Any | None = None
+        raw_content: bytes | None = None
+        if isinstance(body, JSONBody):
+            json_data = body.data
+        elif isinstance(body, BytesBody):
+            raw_content = body.data
+            request_headers["Content-Type"] = body.content_type
```
1st1, line 105: This block is 100% duplicated in `AsyncTransport` -- move it into a shared helper function. Baisically the shared function should configure the `httpx.Client` or `htttpx.AsyncClient` instance (read it from `self.__HTTPX_CLIENT_CLASS__`)

src/vercel/blob/ops.py
```diff
@@ -10,6 +10,13 @@
 import httpx
 
 from .._telemetry.tracker import telemetry, track
+from ._core import (
+    build_create_folder_result,
+    build_head_blob_result,
+    build_list_blob_result,
+    build_list_params,
+    build_put_blob_result,
+)
```
1st1, line 19: There's still insane amount of duplicated code in this file

scotttrinh:
Closing in favor of a different starting point

PR #46, opened 2026-02-06
Add regression test suite

scotttrinh:
This is extracted from #44 since we might actually go a different way, but this regression test suite is a nice thing to keep and base other attempts at no-op refactors on.

### DIFF
--- a/src/vercel/projects/projects.py
+++ b/src/vercel/projects/projects.py
@@ -269,7 +269,7 @@ def update_project(
     team_id: str | None = None,
     slug: str | None = None,
     base_url: str = DEFAULT_API_BASE_URL,
-    timeout: float = 30.0,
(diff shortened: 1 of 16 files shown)


src/vercel/projects/projects.py
```diff
@@ -269,7 +269,7 @@ def update_project(
     team_id: str | None = None,
     slug: str | None = None,
     base_url: str = DEFAULT_API_BASE_URL,
-    timeout: float = 30.0,
+    timeout: float = DEFAULT_TIMEOUT,
```
scotttrinh, line 272: This is an actual behavioral change that is intentional. Someone flag if there is a reason why the async version uses `DEFAULT_TIMEOUT` (which is 60 seconds) and the sync version uses 30 seconds.

scotttrinh:
Closing for now as it's part of #48

PR #48, opened 2026-02-10
Convert `blob` to use `iter_coroutine` and `_http`

scotttrinh:
This PR refactors the Blob module to eliminate sync/async code duplication. All business logic is written once as async and shared between both API surfaces using `iter_coroutine()` — a lightweight bridge that drives non-suspending coroutines synchronously.

  ### What changed

  - Moved all internal implementation into `src/vercel/_internal/`:
    - `_internal/blob/core.py` — request execution (headers, retries, error mapping, progress callbacks, response decoding) and blob operations (`put`, `delete`, `head`, `get`, `list`, `copy`, `create_folder`, `download_file`, `upload_file`) in a `BaseBlobOpsClient` hierarchy
    - `_internal/blob/multipart.py` — multipart upload orchestration with runtime-specific executors: `_SyncMultipartUploadRuntime` (ThreadPoolExecutor) and `_AsyncMultipartUploadRuntime` (anyio task groups)
    - `_internal/http/` — transport abstraction (`SyncTransport`/`AsyncTransport`) wrapping httpx clients with a shared async interface, using `build_request` + `send` for consistent `stream`/`follow_redirects` handling; includes `RawBody` for pass-through streaming bodies
    - `_internal/iter_coroutine.py` — the sync/async bridge that executes non-suspending coroutines in a single `send(None)` step
    - `_internal/telemetry/` — relocated from `_telemetry/`
  - Refactored public API layers to be thin wrappers:
    - `blob/ops.py` — module-level functions (`put`/`put_async`, `delete`/`delete_async`, etc.) delegate to `SyncBlobOpsClient`/`AsyncBlobOpsClient`
    - `blob/client.py` — `BlobClient`/`AsyncBlobClient` manage a token and client lifecycle via context managers
    - `blob/multipart/api.py` — `MultipartUploader`/`AsyncMultipartUploader` wrap multipart create/upload/complete flows
  - Removed `blob/api.py` (duplicated sync+async request logic) and `blob/multipart/core.py` (old multipart implementation)
  - Expanded public exports: added `BlobInvalidResponseJSONError`, `BlobNoTokenProvidedError`, `BlobUnexpectedResponseContentTypeError`, and moved `Access`/`OnUploadProgressCallback`/`UploadProgressEvent` from the deleted `utils.py` into `blob/types.py`

  ### Call flow

Here is the flow for the `put` op as an example:

```
  put(path, body, token=...)                              # ops.py
    → SyncBlobOpsClient (SyncTransport + httpx.Client)    # _internal/blob/core.py
      → iter_coroutine(client.put_blob(...))
        → BaseBlobOpsClient.put_blob (async)
          → small body: request_api("PUT", ...) → transport.send()
          → large body: multipart create → ThreadPoolExecutor upload parts → complete
        → returns PutBlobResult

  put_async(path, body, token=...)                        # ops.py
    → AsyncBlobOpsClient (AsyncTransport + httpx.AsyncClient)
      → await client.put_blob(...)                        # same async code path
```


1st1:
Alright, overall the approach works and IMO this PR is destined to be landed.

Notes:

* `blob.ops.download_file` and `download_file_async` need to be unified
* ditto `blob.ops.upload_file` and `blob.ops.get`

src/vercel/blob/multipart/uploader.py
```diff
@@ -106,6 +171,142 @@ async def _aiter_part_bytes(body: Any, part_size: int) -> AsyncIterator[bytes]:
...
+            )
+            return _normalize_part_upload_result(part_number, response)
+
+        inflight: set[asyncio.Task[dict[str, Any]]] = set()
+        part_number = 1
+        async for chunk in _aiter_part_bytes(body, part_size):
+            task = asyncio.create_task(upload_one(part_number, chunk))
+            inflight.add(task)
+            part_number += 1
+            if len(inflight) >= MAX_CONCURRENCY:
```
fantix: Seems to me that this can be an `asyncio.TaskGroup` + `asyncio.BoundedSemaphore`?

fantix (approved):
Great job! I didn't verify each Vercel API implementation, but the new framework here looks solid to me!

src/vercel/_internal/blob/core.py
```diff
@@ -0,0 +1,1123 @@
...
+
+
+def decode_blob_response(response: httpx.Response) -> Any:
+    content_type = response.headers.get("content-type", "")
+    if "application/json" in content_type or (response.text or "").startswith("{"):
+        try:
+            return response.json()
+        except Exception:
+            return response.text
+    try:
```
elprans, line 163: This is a bit confusing.  Why are we trying to decode json again?

src/vercel/blob/multipart/core.py
```diff
@@ -4,38 +4,138 @@
...
+            ),
+            params={"pathname": path},
+            body=parts,
+        )
+        return cast(dict[str, Any], response)
+
+
+class _SyncMultipartClient(_BaseMultipartClient):
+    async def _request_api(self, **kwargs: Any) -> Any:
+        return request_api(**kwargs)
```
elprans: `request_api` creates a new client/connection every time, which is a potential perf issue.  I think it's preexisting, but this would be a good opportunity to fix this.   Maybe inherit `_[A]syncMultipartClient` from `_[A]syncBlobOpsClient` directly?

src/vercel/blob/multipart/uploader.py
```diff
@@ -106,6 +171,142 @@ async def _aiter_part_bytes(body: Any, part_size: int) -> AsyncIterator[bytes]:
...
+
+        def upload_one(part_number: int, content: bytes) -> dict[str, Any]:
+            def progress(evt: UploadProgressEvent) -> None:
+                with loaded_lock:
+                    loaded_per_part[part_number] = int(evt.loaded)
+                    if on_upload_progress:
+                        loaded = sum(loaded_per_part.values())
+                        on_upload_progress(_aggregate_progress_event(loaded=loaded, total=total))
+
+            response = call_upload_part(
```
elprans: Likewise, this creates a `MultipartClient` on every iteration, I suggest we look into making the client itself concurrency-safe (via contextvars) and collapse this nesting doll a bit.

src/vercel/blob/multipart/uploader.py
scotttrinh: Good idea! I think we want to use AnyIO instead of asyncio directly, so maybe I'll use this as an opportunity to convert this to use AnyIO? Otherwise, if we want to use `TaskGroup`, we'll need to bump the minimum Python version from the current 3.10 to 3.11. I _think_ AnyIO will work in 3.10 still? Will investigate.

src/vercel/_internal/blob/core.py
scotttrinh: This was the existing pattern that was extracted into this helper function:

https://github.com/vercel/vercel-py/blob/9665242607b4b999a5836810b2033bdfab535a00/src/vercel/blob/api.py#L171-L179

But, I agree that we can do something more sensible here like just attempt to parse the JSON and fallback to text if it fails, or maybe even raise? I'll trace through where this is actually used and see what the downstream effects would be of doing something different here.

src/vercel/blob/multipart/core.py
scotttrinh: Yeah, this was existing, and I was trying to have a light touch. Let me investigate this a bit and see if I can thread it through as you're suggesting here.

src/vercel/blob/multipart/uploader.py
scotttrinh: @fantix ⚡ https://github.com/vercel/vercel-py/pull/48/commits/2d9be69b03d47cd12d595ea1181512b92a7052e3

[Looks like AnyIO works fine for 3.10](https://github.com/agronholm/anyio/actions/runs/22257254185/job/64389929828) with their `TaskGroup`, so did a refactor. wdyt?

fantix: That's very cool! Would `async with semaphore:` work on 3.10 too?

src/vercel/_internal/blob/core.py
scotttrinh: @elprans ⚡ https://github.com/vercel/vercel-py/pull/48/commits/1b063bcc4cb12b816330898b42fe3ff5dc5a7917

This is what I came up with. It's a slight breaking change since we now raise a more specific exception instead of some downstream consumer treating the `str` as a `dict` and getting  a `KeyError` or something like that. wdyt? Should I revert to keeping the same behavior but only do the single parse attempt?

src/vercel/blob/multipart/uploader.py
scotttrinh: It doesn't seem like `Semaphore` is a context manager: https://anyio.readthedocs.io/en/stable/api.html#anyio.Semaphore

I was thinking the same thing though and took a few swings at trying to get the acquire/release flow to feel more idiomatic and this is as close as I could get with me and my helpful AI code assistant.

src/vercel/blob/multipart/core.py
scotttrinh: OK, I _think_ this addresses most of this: https://github.com/vercel/vercel-py/pull/48/commits/ef9db2aa0226b012a6a37f6643b9ab6fb4975a3b

I didn't go contextvar due to concerns about sync behavior on worker threads, but I can revisit if you want me to update it. Ideally we have something _way_ more ergonomic once we revisit the public API in a breaking change in the future.

src/vercel/_internal/blob/core.py
elprans: Inherit `BlobInvalidResponseJSONError` from `ValueError` if you're concerned with maintaining maxium backwards compat.  KeyError et al I don't think we should preserve.

elprans:
@scotttrinh , should we move to the `_internal` package approach to cleanly delineate internals from public API?

src/vercel/blob/multipart/core.py
scotttrinh: Looks like I didn't totally finish the play here, but https://github.com/vercel/vercel-py/pull/48/commits/b877b55863bc5abe1037f7d58b1c2aa43afb69b1 is about as far as I could take it. The standalone functions still instantiate a new client/transport every time, but hopefully we can discourage the usage there.

scotttrinh:
@elprans 

> @scotttrinh , should we move to the _internal package approach to cleanly delineate internals from public API?

I'm ok with that. The "public" API is pretty broad right now, but all of these new `_` modules could move into an `_internal` package, sure! What does that look like? Would each submodule (like `vercel.blob` and `vercel.blob.multipart`) have an `_internal` module within it?

ricardo-agz (approved):

PR #50, opened 2026-02-17
feat: add private blob storage support

vvo:
## Summary

- All blob operations (`put`, `copy`, `get`, multipart) now accept `access="private"` in addition to `"public"`
- Private blobs use `*.private.blob.vercel-storage.com` domains and authenticate via Bearer token
- New `GetBlobResult` type with `.content`, `.etag`, `.size`, `.content_type`, etc.
- Sends `x-vercel-blob-access` header to the API on all mutating operations

## Breaking changes

- **`get()` / `get_async()`** return `GetBlobResult` instead of `bytes`. Migrate: `get(url)` → `get(url).content`
- **`get()` / `get_async()`** now require `access` param (defaults to `"public"`, backward-compatible for public stores)
- New params on `get`: `use_cache`, `if_none_match` (for conditional 304 requests)
- `download_file` / `download_file_async` gain an `access` param (defaults to `"public"`)

## Test plan

- [ ] Verify `put()` + `get()` with `access="private"` against a private blob store
- [ ] Verify `put()` + `get()` with `access="public"` still works (regression)
- [ ] Verify multipart upload with `access="private"`
- [ ] Verify `download_file()` with `access="private"`
- [ ] Verify `BlobNotFoundError` still raised on 404


ricardo-agz (approved):
Nice!

PR #60, opened 2026-03-12
Add core of the new API design

scotttrinh:
This begins the clean-room `vercel.stable` API redesign: a lazy root client now owns transport and lifecycle, and service clients become thin views over that shared runtime instead of each service building its own HTTP/auth stack.

In this slice, the stable root surface lands with SDK (`projects`, `deployments`) and cache clients, plus shared stable errors/options/pagination. The generic `RequestClient` is reduced to transport concerns only, while SDK and cache-specific request policy moves into stable internal clients. Sync and async stay aligned by sharing the same internal backend shape and transport lineage, with tests focused on lazy env resolution, transport reuse, option overlays, and stable error behavior.

## Reviewer notes

Focus on whether the new ownership boundary is right:
- root client owns lazy transport initialization and shutdown
- derived clients reuse that transport but get isolated request state for `with_options()` overlays
- service-specific auth/config/error handling now lives above the generic HTTP layer

## Edits

- Decided to remove the cache refactor until I have a better understanding of how the python runtime use of the `context` ContextVar.


src/vercel/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Top-level exports for the clean-room stable Vercel client surface."""
+
+from vercel.stable.client import AsyncVercel, SyncVercel, create_async_client, create_sync_client
```
fantix, line 3: I think `SyncVercel` is not a very popular naming convention in Python? It's more like `VercelClient` for blocking I/O client.

src/vercel/stable/sdk/client.py
```diff
@@ -0,0 +1,152 @@
...
+    SdkRequestState,
+    create_async_request_client,
+    create_sync_request_client,
+)
+from vercel.stable.options import SdkOptions
+from vercel.stable.sdk.deployments import AsyncDeploymentsClient, SyncDeploymentsClient
+from vercel.stable.sdk.projects import AsyncProjectsClient, SyncProjectsClient
+
+
+@dataclass(frozen=True, slots=True)
```
fantix, line 22: I think the client-level objects should just be regular Python objects instead of dataclasses - it looks like we only use the dataclass-generated `__init__` function, and client objects shouldn't be used in e.g.` dataclasses.asdict(client.get_sdk)`?

src/vercel/_internal/stable/sdk/projects.py
```diff
@@ -0,0 +1,257 @@
...
+            return None
+        return page
+
+
+def _project_path(id_or_name: str) -> str:
+    quoted = urllib.parse.quote(id_or_name, safe="")
+    return f"/v9/projects/{quoted}"
+
+
+def _project_write_body(
```
fantix, line 170: This comment is more like a reminder for myself to discuss: do we want a consistent way to validate user inputs? Or rather, depend on Python typing system and assume all is good, letting the server raise readable errors. For now I can see both up and down sides: server has more accurate and up-to-date APIs, but at the same time, the lack of client-side verification may lead to undefined behaviors like, expecting a string but passed in an integer, succeeds in a code path that is unusual and unexpected.

src/vercel/_internal/stable/sdk/projects.py
```diff
@@ -0,0 +1,257 @@
...
+        name=str(payload["name"]),
+        framework=_optional_str(payload.get("framework")),
+        account_id=_optional_str(payload.get("accountId")),
+        created_at=_optional_int(payload.get("createdAt")),
+        updated_at=_optional_int(payload.get("updatedAt")),
+        raw=dict(payload),
+    )
+
+
+def _parse_project_page(payload: dict[str, Any]) -> ProjectPage:
```
fantix, line 227: In addition to the previous comment on `_project_write_body` - these `_parse_*()` functions could potentially get reused to verify both server outputs and user inputs, like `model_validate()` in Pydantic.

fantix (approved):
I think this is overall great and solid!

src/vercel/stable/options.py
```diff
@@ -0,0 +1,74 @@
...
+import os
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from typing import Any
+
+
+@dataclass(frozen=True, slots=True)
+class RootOptions:
+    timeout: float | None = None
+    env: Mapping[str, str] = field(default_factory=lambda: os.environ)
```
elprans, line 15: Since this is `frozen` you should be taking a copy of `os.environ` instead of a reference to the mutable dict.

src/vercel/stable/client.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def create_sync_client(
+    *,
+    timeout: float | None = None,
+    env: Mapping[str, str] | None = None,
+) -> SyncVercel:
+    return SyncVercel(
+        _runtime=SyncRuntime(),
+        _options=RootOptions(timeout=timeout, env=env if env is not None else os.environ),
```
elprans, line 183: make a copy of passed `env` or `os.environ` here

src/vercel/stable/client.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def create_async_client(
+    *,
+    timeout: float | None = None,
+    env: Mapping[str, str] | None = None,
+) -> AsyncVercel:
+    return AsyncVercel(
+        _runtime=AsyncRuntime(),
+        _options=RootOptions(timeout=timeout, env=env if env is not None else os.environ),
```
elprans, line 194: here too

elprans (approved):

src/vercel/stable/options.py
scotttrinh: Went with a `__post_init__` wrapper of the `env` property which calls `dict` -> `MappingProxyType` since this is read-only. This applies to even explicitly passed in params, not just `os.environ` which feels more robust in general (nothing stopping you from passing `os.environ` is yourself!).

PR #69, opened 2026-03-24
sandbox: Add blocking sandbox stop

scotttrinh:
Adds blocking sandbox stop.

### DIFF
--- a/src/vercel/sandbox/sandbox.py
+++ b/src/vercel/sandbox/sandbox.py
@@ -398,9 +398,9 @@ class AsyncSandbox:
                 ``"stopped"`` within *timeout*.
         """
         response = await self.client.stop_sandbox(sandbox_id=self.sandbox.id)
+        self.sandbox = response.sandbox
         if not blocking:
             return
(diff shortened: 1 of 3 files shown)


src/vercel/sandbox/sandbox.py
```diff
@@ -377,8 +377,31 @@ async def write_files(self, files: builtins.list[WriteFile]) -> None:
...
+                ``blocking=True``.
+
+        Raises:
+            TimeoutError: If ``blocking=True`` and the sandbox does not reach
+                ``"stopped"`` within *timeout*.
+        """
+        response = await self.client.stop_sandbox(sandbox_id=self.sandbox.id)
+        if not blocking:
+            return
+        self.sandbox = response.sandbox
```
ctgowrie: Why do we only mutate if blocking?

scotttrinh: Yeah, good catch, the status is likely changed in the response even without waiting for it to stop fully. Will fix.

ctgowrie (approved):

PR #75, opened 2026-03-27
feat(workflow): add hook support

fantix:
Demo code from docs, workflow:

```python
from vercel.workflow import HookMixin, workflow


# Human approval for AI-generated drafts
class Approval(BaseModel, HookMixin):
    decision: Literal["approved", "changes"]
    notes: str | None


@workflow
async def ai_approval_workflow(topic: str):
    draft = await generate_draft(topic)

    # Wait for human approval events
    events = Approval.wait(token="draft-123")

    async for event in events:
        if event["decision"] == "approved":
            await publish_draft(draft)
            break
        else:
            revised = await refine_draft(draft, event.get("notes"))
            await publish_draft(revised)
            break
```

App:

```python
@app.post("/api/resume")
async def resume_workflow(approval: Approval):
    await approval.resume("draft-123")

    return {"ok": True}
```


src/vercel/workflow/core.py
```diff
@@ -65,3 +80,83 @@ async def sleep(param: int | float | datetime.datetime | str) -> None:
...
+            obj = dataclasses.asdict(self, dict_factory=kwargs.pop("dict_factory", dict))
+            json_str = json.dumps(obj, **kwargs)
+        else:
+            raise TypeError("resume only supports pydantic models or dataclasses")
+
+        return await runtime.resume_hook(token_or_hook, json_str)
+
+
+# must not import in workflow
+# celery/temporal/CLI style, how to import subapps
```
scotttrinh: I assume this note needs to move somewhere more durable or as a follow-up ticket?

tests/test_nanoid.py
```diff
@@ -0,0 +1,174 @@
+"""Test nanoid implementation."""
```
scotttrinh, line 1: Not a blocker, but we have hypothesis installed, and a lot of these unit tests could be expressed as a few property-based tests and get extended coverage for "free" and more directly encode the invariant.

scotttrinh (approved):
It would be good to add an example or two which get run on CI as tests, but we can do that as a follow-up as well.

src/vercel/workflow/core.py
fantix: Ahh this was my notes, I thought I dropped them ... maybe not in this branch! Thanks!

PR #78, opened 2026-03-30
sandbox: Take `timedelta` for expiration

scotttrinh:
Accept `timedelta` for sandbox snapshot expiration and timeout-related APIs

### DIFF
--- a/src/vercel/_internal/sandbox/time.py
+++ b/src/vercel/_internal/sandbox/time.py
@@ -1,0 +1,13 @@
+from __future__ import annotations
+
+from datetime import timedelta
+
+
+def normalize_duration_ms(value: int | timedelta | None) -> int | None:
+    match value:
+        case None:
+            return None
+        case timedelta():
+            return int(value.total_seconds() * 1000)
+        case _:
+            return int(value)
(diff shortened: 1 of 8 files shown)


src/vercel/_internal/sandbox/time.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def normalize_duration_ms(value: int | timedelta | None) -> int | None:
+    match value:
+        case None:
+            return None
+        case timedelta():
+            return int(value.total_seconds() * 1000)
+        case _:
+            return int(value)
```
1st1: This would also accept strings and any object that has `__int__` -- do we have to be permissive here?

1st1 (approved):

src/vercel/_internal/sandbox/time.py
scotttrinh: Fixed! Also found some issues with using `total_seconds() * 1000` since that has some float -> int precision issues. The proper way to get milliseconds of a `timedelta` is `value // timedelta(milliseconds=1)` as per the docs: https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds

> For interval units other than seconds, use the division form directly (for example, `td / timedelta(microseconds=1)`).

PR #79, opened 2026-03-31
sandbox: Simplify Sandbox/Snapshot pagination

scotttrinh:
Add direct iteration support to sandbox and snapshot page objects. This iterates only the items already loaded on that page. Leave iter_items() behavior unchanged for now so cross-page iteration still uses the existing explicit API.

### DIFF
--- a/examples/sandbox_11_snapshots.py
+++ b/examples/sandbox_11_snapshots.py
@@ -131,12 +131,14 @@ async def async_demo() -> None:
 
     paged_ids: list[list[str]] = []
     found_ids: set[str] = set()
-    async for page in pager.iter_pages():
+    page = first_page
+    while page is not None:
         page_ids = [listed.id for listed in page.snapshots]
         paged_ids.append(page_ids)
         found_ids.update(page_ids)
         if all(snapshot_id in found_ids for snapshot_id in async_snapshot_ids):
             break
+        page = await page.get_next_page()
 
     print(f"    Visited pages: {paged_ids}")
     assert all(snapshot_id in found_ids for snapshot_id in async_snapshot_ids), (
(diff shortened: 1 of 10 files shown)


src/vercel/sandbox/page.py
```diff
@@ -58,6 +58,9 @@ def iter_pages(self) -> Iterator[SandboxPage]:
     def iter_items(self) -> Iterator[SandboxModel]:
         return self._controller.iter_items_sync(self)
 
+    def __iter__(self) -> Iterator[SandboxModel]:
+        return iter(self.sandboxes)
```
1st1: I'm confused. `iter_items` supposed to do the same that `__iter__` you add but implementation is different? Why?

Also can we drop `iter_items()`? The name is very bad, in Python std lib it's reserved for iterating over key/values, here we're iterating over just values. This is bad api.

examples/sandbox_11_snapshots.py
```diff
@@ -131,23 +131,27 @@ async def async_demo() -> None:
 
     paged_ids: list[list[str]] = []
     found_ids: set[str] = set()
-    async for page in pager.iter_pages():
+    page = first_page
+    while page is not None:
```
1st1, line 135:
```
async for snapshot in AsyncSnapshot.list(limit=10, since=since):
```

examples/sandbox_11_snapshots.py
```diff
@@ -252,23 +256,27 @@ def sync_demo() -> None:
...
 
     print(f"    Visited pages: {paged_ids}")
     assert all(snapshot_id in found_ids for snapshot_id in sync_snapshot_ids), (
         "Did not find all sync snapshots in Snapshot.list() pages"
     )
 
     item_ids: list[str] = []
-    for listed in Snapshot.list(limit=10, since=since).iter_items():
-        item_ids.append(listed.id)
+    page = Snapshot.list(limit=10, since=since)
```
1st1, line 274:
```
for snapshot in Snapshot.list(limit=10, since=since):
```

1st1 (commented):
Let's not expose `pages`, let's rather implement direct `__aiter__` and `__iter__` on "list" objects. Cleaner API this way.

scotttrinh:
Closing in preference of doing the new idea (no pages) on a clean PR against `main` instead of updating this one.

PR #81, opened 2026-04-07
sandbox: Add typed create payloads

scotttrinh:
This PR tightens the sandbox create surface so nested inputs are typed and validated locally instead of being treated as loose dicts. The goal is to make `source`, `resources`, and network policy inputs more predictable for SDK users while still sending the same backend payload shape.

This branch also captures the implementation iteration we just went through: we first introduced a hand-rolled parsing/validation layer to prove out the typed create boundary, then replaced that machinery with Pydantic models once it was clear the problem space wanted stronger structure and less custom serialization logic.

The result is better Python types, clearer validation errors before requests are sent, and a more maintainable foundation for nested sandbox resource models.


.github/workflows/ci.yml
```diff
@@ -23,11 +23,8 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v3
 
-      - name: Set up virtualenv (Python 3.12)
-        run: uv venv --python 3.12
-
-      - name: Install dependencies
-        run: uv pip install -e .[dev]
+      - name: Sync dependencies
+        run: uv sync --python 3.12
```
scotttrinh, line 27: We had sudden CI failures when it wasn't failing locally, and while trying to troubleshoot this, determined that it was due to not actually using the same `ruff` version locally and in CI due to our explicit ignoring of `uv.lock`. I kept this simplification from that exploration, but it's not strictly necessary.

src/vercel/_internal/sandbox/core.py
```diff
@@ -206,6 +208,40 @@ def _build_sandbox_error(
...
+    if source is not None:
+        body["source"] = source
+    if timeout is not None:
+        body["timeout"] = timeout
+    if resources is not None:
+        body["resources"] = resources
+    if runtime is not None:
+        body["runtime"] = runtime
+    if network_policy is not None:
+        body["network_policy"] = ApiNetworkPolicy.from_network_policy(network_policy).to_dict()
```
scotttrinh: Considered moving this to a `to_payload` structure, but decided that this was out-of-scope for this change due to the complexity of the base `NetworkPayload` type (which is a union of string literals and a dict-like dataclass). In the future, I think it'd be good to actually have a proper type that participates in the `to_payload` protocol.

src/vercel/_internal/sandbox/core.py
```diff
@@ -206,6 +208,40 @@ def _build_sandbox_error(
...
+    if interactive:
+        body["__interactive"] = True
+    if env is not None:
+        body["env"] = RawPayload(env)
```
scotttrinh: tl;dr `RawPayload` skips the snake_case to camelCase conversion that `marshal_payload` does. we don't want to change the keys of `env` since they're environment variables.

src/vercel/_internal/sandbox/models.py
```diff
@@ -20,37 +23,282 @@ class SandboxStatus(StrEnum):
...
+            payload["vcpus"] = self.vcpus
+        if self.memory is not None:
+            payload["memory"] = self.memory
+        return payload
+
+
+ResourcesInput: TypeAlias = Resources | Mapping[str, Any]
+
+
+def parse_source(value: SourceInput | None) -> Source | None:
```
scotttrinh, line 490: We're making these standalone functions instead of static methods on the `Source` class to allow parsing both `Mapping` and existing `Source` values through the same validator. We could instead use `from_dict` on `Source` and skip validating explicit `Source` instances.

tests/integration/test_sandbox_sync_async.py
```diff
@@ -138,6 +142,75 @@ async def test_create_sandbox_async(self, mock_env_clear, mock_sandbox_create_re
...
+                    "routes": [],
+                },
+            )
+        )
+
+        sync_sandbox = Sandbox.create(
+            token="test_token",
+            team_id="team_test123",
+            project_id="prj_test123",
+            source={"type": "git", "url": "https://github.com/vercel/vercel-py"},
```
1st1, line 165: I'd deprecate this -- poor typing and DX

src/vercel/_internal/sandbox/models.py
```diff
@@ -20,37 +23,282 @@ class SandboxStatus(StrEnum):
...
 
 
-class GitSource(_GitSourceRequired, total=False):
-    """Git repository source for creating a sandbox."""
+class SandboxValidationError(SandboxError):
+    """Local sandbox input validation failed before the API request was sent."""
+
+    def __init__(self, issues: list[SandboxValidationIssue]) -> None:
+        self.issues = tuple(issues)
+        message = "; ".join(f"{issue.path}: {issue.message}" for issue in self.issues)
```
1st1:
```suggestion
        message = ";\n\n".join(f"{issue.path}: {issue.message}" for issue in self.issues)
```

1st1 (approved):

PR #83, opened 2026-04-08
Make sandbox lists item-first

scotttrinh:
This change makes sandbox and snapshot listing item-first across the public API by having Sandbox.list() and AsyncSandbox.list() yield models directly, with limit treated as a total yielded-item cap across internal pagination. It removes the public page module and page-oriented exports and helpers, updates the examples and sandbox list coverage to teach direct iteration, and keeps any remaining pagination mechanics as internal implementation details rather than consumer-facing concepts.


src/vercel/sandbox/sandbox.py
```diff
@@ -287,9 +196,36 @@ def list(
...
             until=until,
         )
-        return AsyncSandboxPager(
-            _fetch_first_page=lambda: _build_async_sandbox_page(creds=creds, params=params)
-        )
+
+        async def iter_sandboxes() -> AsyncIterator[SandboxModel]:
+            current_params = params
+            while True:
+                client = AsyncSandboxOpsClient(team_id=creds.team_id, token=creds.token)
```
1st1: Do you want to re-instantiate the client on every iteration?

src/vercel/sandbox/sandbox.py
```diff
@@ -287,9 +196,36 @@ def list(
...
+                for sandbox in sandboxes:
+                    yield sandbox
+                if response.pagination.next is None:
+                    return
+                if (
+                    current_params.remaining is not None
+                    and len(sandboxes) >= current_params.remaining
+                ):
+                    return
+                current_params = current_params.with_until(
```
1st1: Are you advancing `current_params.since`?

1st1: I don't see that in with_until

1st1: alright, i think we're actually advancing `until` so this is ok

1st1 (approved):

PR #85, opened 2026-04-09
Add sandbox request config

scotttrinh:
## Summary
- add a public SandboxRequestConfig surface for sandbox create/get/list APIs
- thread timeout and retry settings into sync and async sandbox ops clients
- document the contract and add workflow coverage showing request defaults persist without changing wait semantics

## Motivation
Sandbox callers could not tune transport-level timeout or retry behavior without wrapping the public API. This exposes that configuration directly while keeping workflow-level wait controls separate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,33 @@ pip install vercel
 
 This package provides both synchronous and asynchronous clients to interact with the Vercel API.
 
+### Sandbox Request Configuration
+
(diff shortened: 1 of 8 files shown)


scotttrinh:
Closing this. What we really want is a "client" interface here that we can configure, and we'll be tackling that as part of the V2 move soon.

PR #89, opened 2026-04-10
Add support for python 3.10

scotttrinh:
We recently bumped to 3.11 as our minimum version, but this restores the ability to use 3.10 with some shims until EOL in October.

### DIFF
--- a/src/vercel/_internal/sandbox/models.py
+++ b/src/vercel/_internal/sandbox/models.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+import sys
 from collections.abc import Mapping, Sequence
 from datetime import timedelta
-from enum import StrEnum
+from enum import Enum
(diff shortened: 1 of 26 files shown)


src/vercel/_internal/sandbox/models.py
```diff
@@ -26,8 +27,18 @@
 # Source types for Sandbox.create()
 _REDACTED_HEADER_VALUE = "<redacted>"
 
+if sys.version_info >= (3, 11):
+    from enum import StrEnum as _StrEnum
+else:
 
-class SandboxStatus(StrEnum):
+    class _StrEnum(str, Enum):  # noqa: UP042
```
elprans: A more complete polyfill: https://github.com/geldata/gel-python/blob/8a956d62b380db36ecaacc57952a94b1bab40d6a/gel/_internal/_polyfills/_strenum_impl.py#L15

elprans (approved):

PR #90, opened 2026-04-10
Refine sandbox duration coercion with typed and boundary helpers

### LINKED ISSUE #86: Use timedelta across sandbox duration APIs
## Summary

Sandbox duration APIs are inconsistent. Some public inputs already accept `timedelta`, but the lifecycle wait APIs still require raw second floats, and the public models still expose raw millisecond integers.

## Current State

- `Sandbox.create(timeout=...)` and `AsyncSandbox.create(timeout=...)` accept `int | timedelta`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L123-L177
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L551-L608
- `extend_timeout(duration=...)` accepts `int | timedelta`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L439-L451
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L877-L891
- `snapshot(expiration=...)` accepts `int | timedelta | SnapshotExpiration | None`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L453-L468
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L893-L910
- The shared internal normalizer already exists:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/_internal/sandbox/time.py#L1-L15

But the lifecycle wait APIs still use raw second floats:

- `AsyncSandbox.wait_for_status(timeout: float, poll_interval: float)`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L280-L310
- `Sandbox.wait_for_status(timeout: float, poll_interval: float)`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L714-L742
- `AsyncSandbox.stop(blocking=True, timeout: float, poll_interval: float)`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L413-L437
- `Sandbox.stop(blocking=True, timeout: float, poll_interval: float)`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L851-L875

The public model and wrapper properties also still expose raw millisecond integers:

- `AsyncSandbox.timeout` / `Sandbox.timeout`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L106-L109
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/sandbox/sandbox.py#L542-L545
- Internal public-facing model fields `timeout: int` and `duration: int | None`:
  - https://github.com/vercel/vercel-py/blob/main/src/vercel/_internal/sandbox/models.py#L542-L587

## Desired Direction

- Treat `timedelta` as the first-class public representation for timeout and duration values.
- Keep `timedelta` internally until values need to be converted to the API's numeric literal format.
- Limit integer conversions to API boundary code.

## Scope

- Update `wait_for_status()` and `stop(blocking=True)` to accept `timedelta`-style public inputs.
- Audit whether public timeout/duration properties should expose `timedelta` instead of raw millisecond integers.
- Reuse a shared duration codec instead of duplicating conversions in call sites.

scotttrinh:
This PR cleans up the internal duration-conversion API used by the sandbox client.

The main outcome is a deliberate split between two helpers in `src/vercel/_internal/sandbox/time.py`:

- `coerce_duration(value, unit)` for typed internal call sites
- `parse_duration(value, unit)` for untyped boundary inputs such as Pydantic field validators

It also exports unit constants:

- `MILLISECOND`
- `SECOND`

This keeps the core conversion logic generic without continuing to grow a family of unit-specific helpers.

## Current API

The current shape is:

```python
delta = coerce_duration(timeout, SECOND)
seconds = delta / SECOND
```

```python
delta = coerce_duration(duration, MILLISECOND)
milliseconds = delta // MILLISECOND
```

For untyped boundary inputs:

```python
normalized_delta = parse_duration(value, MILLISECOND)
if normalized_delta is None:
    return None
return normalized_delta // MILLISECOND
```

This makes two things explicit:

- coercion into a canonical internal representation (`timedelta`)
- projection back into the numeric form the caller actually wants

## Why split `coerce_duration` and `parse_duration`

`coerce_duration()` is intentionally typed:

```python
def coerce_duration(value: int | float | timedelta, unit: timedelta) -> timedelta
```

That makes internal call sites crisp and keeps type checking useful.

`parse_duration()` is intentionally boundary-oriented:

```python
def parse_duration(value: object, unit: timedelta) -> timedelta | None
```

That fits places like Pydantic validators, where the input is genuinely `object` and may be invalid or `None`.

This avoids pushing boundary-validation concerns into the typed helper.

## None handling

We explored letting `coerce_duration()` accept `None`, but it made the ergonomics worse for normal internal call sites:

- it forced optional return types into places that did not need them
- it required extra guards or assertions before applying `/ SECOND` or `// MILLISECOND`

The final decision was:

- `coerce_duration()` does not handle `None`
- `parse_duration()` does handle `None`

That keeps the internal helper simple while still supporting boundary-facing validators cleanly.

### 1. Keep adding more specialized helpers

Example API:

```python
seconds = normalize_duration_seconds(timeout)
milliseconds = normalize_duration_ms(duration)
```

Pros:

- compact call sites
- easy to read in isolation

Cons:

- the helper family keeps growing by unit and output shape
- the real shared abstraction stays hidden
- tests and behavior tend to duplicate across helpers

This is the direction the code was already drifting toward, and this PR intentionally moves away from it.

### 2. Add an `as_type` kwarg to `coerce_duration`

Example API:

```python
seconds = coerce_duration(timeout, SECOND, as_type=float)
milliseconds = coerce_duration(duration, MILLISECOND, as_type=int)
```

Pros:

- one entry point
- hides the projection step

Cons:

- mixes coercion and numeric projection again
- makes the helper responsible for deciding between `/` and `//`
- gets awkward once `timedelta` inputs are involved, because the intended output shape is policy, not something recoverable from the input

This felt more clever than clear.

### 3. Accept a `project` or `transform` lambda

Example API:

```python
seconds = coerce_duration(timeout, SECOND, project=lambda d: d / SECOND)
milliseconds = coerce_duration(duration, MILLISECOND, project=lambda d: d // MILLISECOND)
```

Pros:

- flexible
- can hide optional-style mapping behavior inside the helper

Cons:

- pushes more abstraction into a helper that should stay simple
- makes straightforward unit conversion harder to scan at the call site
- effectively reintroduces “option map” behavior without much Pythonic benefit

We decided this was possible, but not better.

## Why the current shape won

The chosen API keeps the split simple:

- `coerce_duration()` answers “how do I turn a typed duration input into a `timedelta`?”
- `parse_duration()` answers “how do I safely accept an untyped boundary value and maybe get a `timedelta`?”

Then each caller makes its numeric projection explicit:

- `/ SECOND` for fractional seconds
- `// MILLISECOND` for integer milliseconds

That keeps unit semantics visible at the call site and avoids making one helper carry too much policy.

Closes #86

### DIFF
--- a/src/vercel/_internal/sandbox/core.py
+++ b/src/vercel/_internal/sandbox/core.py
@@ -63,7 +63,7 @@ from vercel._internal.sandbox.models import (
     WriteFile,
 )
 from vercel._internal.sandbox.snapshot import SnapshotExpiration
-from vercel._internal.sandbox.time import normalize_duration_ms
+from vercel._internal.sandbox.time import MILLISECOND, coerce_duration
 
 try:
     VERSION = _pkg_version("vercel")
@@ -508,7 +508,7 @@ class BaseSandboxOpsClient:
         data = await self._request_client.request_json(
             "POST",
             f"/v1/sandboxes/{sandbox_id}/extend-timeout",
-            body=JSONBody({"duration": normalize_duration_ms(duration)}),
+            body=JSONBody({"duration": coerce_duration(duration, MILLISECOND) // MILLISECOND}),
         )
         return SandboxResponse.model_validate(data)
 
--- a/src/vercel/_internal/sandbox/core.py
+++ b/src/vercel/_internal/sandbox/core.py
@@ -63,7 +63,7 @@ from vercel._internal.sandbox.models import (
     WriteFile,
 )
 from vercel._internal.sandbox.snapshot import SnapshotExpiration
-from vercel._internal.sandbox.time import MILLISECOND, coerce_duration
+from vercel._internal.sandbox.time import to_ms_int
 
 try:
     VERSION = _pkg_version("vercel")
@@ -502,13 +502,11 @@ class BaseSandboxOpsClient:
             body=JSONBody({"signal": signal}),
         )
 
-    async def extend_timeout(
-        self, *, sandbox_id: str, duration: int | timedelta
-    ) -> SandboxResponse:
+    async def extend_timeout(self, *, sandbox_id: str, duration: timedelta) -> SandboxResponse:
         data = await self._request_client.request_json(
             "POST",
             f"/v1/sandboxes/{sandbox_id}/extend-timeout",
-            body=JSONBody({"duration": coerce_duration(duration, MILLISECOND) // MILLISECOND}),
+            body=JSONBody({"duration": to_ms_int(duration)}),
         )
         return SandboxResponse.model_validate(data)
 
--- a/src/vercel/_internal/sandbox/snapshot.py
+++ b/src/vercel/_internal/sandbox/snapshot.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from datetime import timedelta
 
-from vercel._internal.sandbox.time import MILLISECOND, coerce_duration
+from vercel._internal.sandbox.time import MILLISECOND, coerce_duration, to_ms_int
 
 MIN_SNAPSHOT_EXPIRATION = timedelta(milliseconds=86_400_000)
 _ZERO_DELTA = timedelta(seconds=0)
@@ -27,7 +27,7 @@ class SnapshotExpiration:
         self._td = normalized_delta
 
     def __int__(self) -> int:
-        return self._td // MILLISECOND
+        return to_ms_int(self._td)
 
     def __eq__(self, other: object) -> bool:
         match other:
--- a/src/vercel/_internal/sandbox/snapshot.py
+++ b/src/vercel/_internal/sandbox/snapshot.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from datetime import timedelta
 from typing import Final
 
-from vercel._internal.sandbox.time import normalize_duration_ms
+from vercel._internal.sandbox.time import MILLISECOND, coerce_duration
 
 MIN_SNAPSHOT_EXPIRATION_MS: Final[int] = 86_400_000
 
@@ -16,9 +16,8 @@ class SnapshotExpiration(int):
     """
 
     def __new__(cls, value: int | timedelta) -> SnapshotExpiration:
-        normalized_value = normalize_duration_ms(value)
-        if normalized_value is None:
-            raise TypeError("Snapshot expiration cannot be None")
+        normalized_delta = coerce_duration(value, MILLISECOND)
+        normalized_value = normalized_delta // MILLISECOND
         if normalized_value != 0 and normalized_value < MIN_SNAPSHOT_EXPIRATION_MS:
             raise ValueError(
                 "Snapshot expiration must be 0 for no expiration or >= 86400000 milliseconds"
--- a/src/vercel/_internal/sandbox/snapshot.py
+++ b/src/vercel/_internal/sandbox/snapshot.py
@@ -1,25 +1,41 @@
 from __future__ import annotations
 
+from dataclasses import dataclass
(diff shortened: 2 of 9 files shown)


src/vercel/_internal/sandbox/core.py
```diff
@@ -508,7 +508,7 @@ async def extend_timeout(
         data = await self._request_client.request_json(
             "POST",
             f"/v1/sandboxes/{sandbox_id}/extend-timeout",
-            body=JSONBody({"duration": normalize_duration_ms(duration)}),
+            body=JSONBody({"duration": coerce_duration(duration, MILLISECOND) // MILLISECOND}),
```
1st1: I'd define a helper `vercel._internal.sandbox.time.to_ms()`

src/vercel/_internal/sandbox/snapshot.py
1st1: In general it's an anti-pattern to override built-in scalar types. Make it a data class. You can make the data class implement `__int__` if you wish

1st1 (approved):
I'll approve this but please address the comments, particularly the subclassing `int` one - we shouldn't do that.

src/vercel/_internal/sandbox/core.py
scotttrinh: What would it take as an input? A `timedelta`?

```py
duration = coerce_duration(value, MILLISECOND)
ms = to_ms(duration)
```

That's not much different from:

```py
duration = timedelta(millisecond=value)
ms = duration // MILLISECOND
```

Although arguably more clear.

If we are trying to wrap the whole thing up, coercion through numeric conversion, I couldn't come up with a unified API that was type safe and didn't just include every combination of input type, what unit that value represents, and the target output type.

1st1: OK I'm a bit lost now... how is

```
duration = timedelta(millisecond=value)
ms = duration // MILLISECOND
```

different from

```
ms = value
```

?

scotttrinh: yeah, fair, this is definitely not the most obvious thing on first read.

The short version is that it's doing boundary normalization. Our public APIs now accept `timedelta` in addition to bare numerics like `int` or `float`, but when you pass a bare `5`, it doesn't tell you whether it's seconds or milliseconds. So the call site provides the unit, we convert to a `timedelta`, and from there everything internally carries its own unit. At the consumer end we convert out to whatever's needed — `td // MILLISECOND` for API payloads, `td / SECOND` for sleep-style calls.

The main alternative would be to normalize in the other direction: convert incoming `timedelta` to a numeric type at the boundary and carry that through. That's a valid design and would be simpler in some of today's call sites, since we always end up at `int` ms or `float` seconds anyway. I just prefer keeping `timedelta` as the internal representation because it means duration values carry their own unit instead of relying on convention.

fwiw I'm cool to add a couple of read-side helpers like `to_ms_int(td)` and `to_seconds_float(td)` since those are really the only two cases we need today. I'll do that and you can tell me what you think.

1st1: This is fine. Just add a short comment.

PR #95, opened 2026-04-13
workflow: fix concurrent sandbox

fantix:
The same lambda function may host multiple workflow/step handler requests at the same time, we need to be able to switch into/out of sandboxes as concurrent asyncio Tasks switches contexts.

### DIFF
--- a/src/vercel/workflow/py_sandbox.py
+++ b/src/vercel/workflow/py_sandbox.py
@@ -44,7 +44,7 @@ class _ModulePolicy:
     allowed: frozenset[str] | None = None
     allow_if: Callable[[str], bool] | None = None
 
-    def post_exec(self, *, proxy: _ProxyModule, module: types.ModuleType, random_seed: str) -> None:
+    def post_exec(self, *, proxy: _ProxyModule, module: types.ModuleType) -> None:
         pass
 
     def resolve_attr(self, name: str, real: types.ModuleType) -> Any:
@@ -534,13 +534,11 @@ class _SandboxFinder(MetaPathFinder):
         passthrough: set[str],
         restrictions: dict[str, _ModulePolicy] | None = None,
         blocked: set[str] | None = None,
-        random_seed: str,
     ) -> None:
         self._host = host_modules
         self._passthrough = passthrough
         self._restrictions = restrictions or {}
         self._blocked = blocked or set()
-        self._random_seed = random_seed
 
     def _is_passthrough(self, name: str) -> bool:
         for prefix in self._passthrough:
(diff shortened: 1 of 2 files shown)


src/vercel/workflow/py_sandbox.py
```diff
@@ -37,9 +44,35 @@ class _ModulePolicy:
...
+    """
+
+    def _wrapper(*_args: Any, **_kwargs: Any) -> Any:
+        if _in_sandbox.get(False):
+            raise SandboxRestrictionError(
+                f"Cannot call {name}() inside a workflow. Workflows must be deterministic."
+            )
+        return real_fn(*_args, **_kwargs)
+
+    _wrapper.__qualname__ = f"<context-restricted {name}>"
```
1st1:
```suggestion
    _wrapper.__qualname__ = f"<workflow-context-restricted {name}>"
```

src/vercel/workflow/py_sandbox.py
```diff
@@ -37,9 +44,35 @@ class _ModulePolicy:
...
+    def post_exec(self, *, proxy: _ProxyModule, module: types.ModuleType) -> None:
         pass
 
+    def resolve_attr(self, name: str, real: types.ModuleType) -> Any:
+        """Resolve an allowed attribute on the real module.
+
+        Called by ``_ProxyModule.__getattr__`` as the final fallback.
+        Subclasses can override to intercept (e.g. per-context random).
+        """
+        return getattr(real, name)
```
1st1, line 56: Should it be `raise NotImplementedError`?

src/vercel/workflow/py_sandbox.py
```diff
@@ -653,27 +723,49 @@ def workflow_sandbox(*, random_seed: str) -> Iterator[None]:
...
-            # Restore original sys.modules contents.
             sys.modules.clear()
             sys.modules.update(orig_modules)
+
+
+@contextmanager
+def workflow_sandbox(*, random_seed: str) -> Iterator[None]:
+    """Activate the workflow sandbox.
+
+    Patches ``sys.modules`` (ref-counted for concurrent sandboxes),
```
1st1: ~The comment is outdated? I don't see this context manager implementing refcouting or even patching sys.modules, it's more about setting up the context for downstream code?~

1st1: nvm!

1st1 (approved):

src/vercel/workflow/py_sandbox.py
fantix: Negative, this is the allowed path to get the actual member (success or `AttributeError`), it's really just inlined `getattr`, but we want to override it for the `random` module/policy.

src/vercel/workflow/py_sandbox.py
fantix: Ohh yes, the refcounting is refactored out to `_sandbox_modules` and `_sandbox_refcount`, let me update the comments.

PR #97, opened 2026-04-14
feat!: replace global registry with Workflows

fantix:
Depends on #95

Instead of:

```python
from vercel.workflow import *

@workflow
async def my_workflow():
    await my_step()
    ...

@step
async def my_step():
    ...
```

w/ `vercel.json`:

```json
{
  "experimentalServices": {
    "my_workflow": {
      "type": "worker",
      "entrypoint": "backend/mypkg/myworkflow.py",
      "topics": ["__wkf_step_", "__wkf_workflow_"]
    }
}
```

Now we are aiming for:

```python
from vercel import workflow

wf = workflow.Workflows()

@wf.workflow
async def my_workflow():
    await my_step()
    ...

@wf.step
async def my_step():
    ...
```

w/ `vercel.json`:

```json
{
  "experimentalServices": {
    "my_workflow": {
      "type": "job",
      "trigger": "workflow",
      "root": "backend",
      "entrypoint": "mypkg.myworkflow:wf"
    }
}
```


src/vercel/_internal/workflow/core.py
```diff
@@ -192,3 +133,32 @@ async def resume(self, token_or_hook: str | w.Hook, **kwargs) -> w.Hook:
...
+            runtime.workflow_entrypoint(self)
+            runtime.step_entrypoint(self)
+
+    def workflow(self, func: Callable[P, Coroutine[Any, Any, T]]) -> Workflow[P, T]:
+        rv = Workflow(func)
+        assert rv.workflow_id not in self._workflows, f"Duplicate workflow ID: {rv.workflow_id}"
+        self._workflows[rv.workflow_id] = rv
+        return rv
+
+    def get_workflow(self, workflow_id: str) -> Workflow[Any, Any]:
```
1st1: Do we need this as a public API?

src/vercel/_internal/workflow/core.py
```diff
@@ -192,3 +133,32 @@ async def resume(self, token_or_hook: str | w.Hook, **kwargs) -> w.Hook:
...
+        return self._workflows[workflow_id]
+
+    def step(self, func: Callable[P, Coroutine[Any, Any, T]]) -> Step[P, T]:
+        rv = Step(func)
+        assert rv.name not in self._steps, f"Duplicate step name: {rv.name}"
+        self._steps[rv.name] = rv
+        return rv
+
+    def get_step(self, step_name: str) -> Step[Any, Any]:
+        return self._steps[step_name]
```
1st1, line 164: ditto

1st1 (approved):

src/vercel/_internal/workflow/core.py
fantix: no we don't!

1st1: hide it then.

PR #98, opened 2026-04-14
Refactor PTY support around AsyncPTYSession

scotttrinh:
This PR refactors PTY support around a reusable low-level `AsyncPTYSession` so PTY access in `vercel-py` is no longer effectively tied to the terminal-taking `AsyncSandbox.shell()` workflow.

The current shape is:

- `AsyncSandbox.shell()` remains the high-level helper for interactive local terminal takeover
- `AsyncSandbox.open_pty()` is the low-level acquisition path for callers that want PTY access without depending on `sys.stdin` / `sys.stdout`
- `AsyncPTYSession` owns the remote PTY lifecycle end-to-end: PTY helper installation, detached PTY server startup, websocket connection, readiness, resize, metadata, and teardown

This keeps the shell UX intact while giving downstream PTY consumers a first-class session abstraction instead of requiring manual bootstrap work.

The result is a clearer separation between terminal UX and PTY session management, plus a lower-level API that is easier to compose in provider-style integrations.

## API Shape

The low-level PTY contract is now centered on `AsyncPTYSession` plus a stream-shaped I/O surface:

```python
async with await sandbox.open_pty(
    ["/bin/bash"],
    cols=100,
    rows=30,
) as session:
    await session.ready()
    await session.resize(100, 30)

    await session.stream.send(b"printf 'PTY_OK\\n'; pwd; exit\\n")

    output = b""
    async for chunk in session.stream:
        output += chunk
        if b"PTY_OK" in output:
            break
```

That is the intended usage now: lifecycle methods like `ready()` and `resize()` stay on the session, while PTY bytes flow through `session.stream`.

A slightly more direct receive loop looks like this:

```python
await session.stream.send(b"echo hello\\n")

buffer = b""
while True:
    chunk = await session.stream.receive()
    buffer += chunk
    if b"hello" in buffer:
        break
```

Closing the stream closes the session, since there is no separate meaningful I/O path once the websocket-backed PTY stream is gone.

### DIFF
--- a/src/vercel/_internal/sandbox/pty_binary.py
+++ b/src/vercel/_internal/sandbox/pty_binary.py
@@ -1,0 +1,72 @@
+"""Internal PTY server binary management helpers."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import httpx
+
+BINARY_BASE_URL = "https://pty-tunnel.labs.vercel.dev"
+
+CACHE_DIR = Path.home() / ".cache" / "vercel-sandbox"
+
+SERVER_BIN_NAME = "vc-interactive-server"
+
+# Sandboxes are currently always Linux x86_64
+DEFAULT_SANDBOX_ARCH = "x86_64"
+
+
+def get_binary_cache_path(arch: str | None = None) -> Path:
+    if arch is None:
+        arch = DEFAULT_SANDBOX_ARCH
+    return CACHE_DIR / f"pty-server-linux-{arch}"
+
+
+def download_binary(arch: str | None = None, *, force: bool = False) -> Path:
+    if arch is None:
+        arch = DEFAULT_SANDBOX_ARCH
+
+    cache_path = get_binary_cache_path(arch)
+
+    if cache_path.exists() and not force:
+        return cache_path
+
+    CACHE_DIR.mkdir(parents=True, exist_ok=True)
+
+    url = f"{BINARY_BASE_URL}/linux-{arch}"
+    with httpx.Client(timeout=60.0, follow_redirects=True) as client:
+        response = client.get(url)
+        response.raise_for_status()
+        cache_path.write_bytes(response.content)
+
+    return cache_path
+
+
+def get_binary_bytes(arch: str | None = None) -> bytes:
+    path = download_binary(arch)
+    return path.read_bytes()
+
+
+async def download_binary_async(arch: str | None = None, *, force: bool = False) -> Path:
+    if arch is None:
+        arch = DEFAULT_SANDBOX_ARCH
+
+    cache_path = get_binary_cache_path(arch)
+
+    if cache_path.exists() and not force:
+        return cache_path
+
+    CACHE_DIR.mkdir(parents=True, exist_ok=True)
+
+    url = f"{BINARY_BASE_URL}/linux-{arch}"
+    async with httpx.AsyncClient(timeout=60.0, follow_redirects=True) as client:
+        response = await client.get(url)
+        response.raise_for_status()
+        cache_path.write_bytes(response.content)
+
+    return cache_path
+
+
+async def get_binary_bytes_async(arch: str | None = None) -> bytes:
+    path = await download_binary_async(arch)
+    return path.read_bytes()
--- a/src/vercel/_internal/sandbox/pty_binary.py
+++ b/src/vercel/_internal/sandbox/pty_binary.py
@@ -6,6 +6,9 @@ from pathlib import Path
 
 import httpx
 
+from .constants import PTY_BINARY_DOWNLOAD_TIMEOUT
+from .time import to_seconds_float
+
 BINARY_BASE_URL = "https://pty-tunnel.labs.vercel.dev"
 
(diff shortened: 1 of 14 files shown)


src/vercel/_internal/sandbox/pty_binary.py
```diff
@@ -0,0 +1,72 @@
...
+
+    cache_path = get_binary_cache_path(arch)
+
+    if cache_path.exists() and not force:
+        return cache_path
+
+    CACHE_DIR.mkdir(parents=True, exist_ok=True)
+
+    url = f"{BINARY_BASE_URL}/linux-{arch}"
+    with httpx.Client(timeout=60.0, follow_redirects=True) as client:
```
1st1: I'm a bit worried about hard-coding timeouts like this. Can we factor out this into a constant and ensure we don't have timeouts hardcoded like this in the code base elsewhere?

Let's have one `DEFAULT_SANDBOX_API_TIMEOUT` and then later we can grep through it and replace with proper client-specific timeout setting when we ship that new API

src/vercel/_internal/sandbox/pty_binary.py
```diff
@@ -0,0 +1,72 @@
...
+import httpx
+
+BINARY_BASE_URL = "https://pty-tunnel.labs.vercel.dev"
+
+CACHE_DIR = Path.home() / ".cache" / "vercel-sandbox"
+
+SERVER_BIN_NAME = "vc-interactive-server"
+
+# Sandboxes are currently always Linux x86_64
+DEFAULT_SANDBOX_ARCH = "x86_64"
```
1st1, line 19: Do we have a nice error if you specify a different arch?

src/vercel/_internal/sandbox/pty_session.py
```diff
@@ -0,0 +1,275 @@
...
+
+async def setup_sandbox_environment(sandbox: AsyncSandbox) -> None:
+    """Install the PTY server binary in the sandbox if not present."""
+    result = await sandbox.run_command("command", ["-v", SERVER_BIN_NAME])
+    if result.exit_code == 0:
+        return
+
+    binary = await get_binary_bytes_async()
+
+    tmp_path = f"/tmp/{SERVER_BIN_NAME}-install"
```
1st1, line 70: Use the `tempfile` module for this, or look up which `os.` function it uses to find the system-dependent `/tmp/` dir. It's not always `tmp`

1st1: Oh... or this is a path within sandbox? If so add a comment (and this becomes basically sandbox API contract :))

src/vercel/_internal/sandbox/pty_session.py
```diff
@@ -0,0 +1,275 @@
...
+            "-c",
+            f'mv "{tmp_path}" /usr/local/bin/{SERVER_BIN_NAME} && '
+            f"chmod +x /usr/local/bin/{SERVER_BIN_NAME}",
+        ],
+        sudo=True,
+    )
+
+
+async def read_connection_info(
+    cmd: AsyncCommand, timeout: DurationSeconds = 30.0
```
1st1: Again, another hardcoded timeouts. We have to have one file with constants for this, or just one constant. Or this will become unmaintainable.

src/vercel/_internal/sandbox/pty_session.py
```diff
@@ -0,0 +1,275 @@
...
+            collected += log.data
+            for line in collected.split("\n"):
+                line = line.strip()
+                if not (line.startswith("{") and line.endswith("}")):
+                    continue
+                try:
+                    return json.loads(line)
+                except json.JSONDecodeError:
+                    continue
+        return None
```
1st1, line 130: This code is quadratic and smells weird to me. For once, it's quadratic, second - you're only parsing one log line discarding the rest of the data which might have more log lines in it.

src/vercel/_internal/sandbox/pty_session.py
```diff
@@ -0,0 +1,275 @@
...
+async def start_pty_server(
+    sandbox: AsyncSandbox,
+    command: list[str],
+    *,
+    env: dict[str, str] | None = None,
+    cwd: str | None = None,
+    sudo: bool = False,
+    cols: int | None = None,
+    rows: int | None = None,
+    connection_timeout: DurationSeconds = 30.0,
```
1st1: ditto - hardcoded timeout

src/vercel/_internal/sandbox/pty_session.py
```diff
@@ -0,0 +1,275 @@
...
+        sandbox: AsyncSandbox,
+        command: list[str] | None = None,
+        *,
+        env: dict[str, str] | None = None,
+        cwd: str | None = None,
+        sudo: bool = False,
+        cols: int | None = None,
+        rows: int | None = None,
+        _client_factory: PTYClientFactory | None = None,
+        _connection_timeout: DurationSeconds = 30.0,
```
1st1: ditto

1st1 (commented):
Left some comments, need to bake this a bit longer

1st1 (approved):

PR #105, opened 2026-04-15
Get OIDC token from request headers in get_credentials

ricardo-agz:
Use get_vercel_oidc_token_from_context() instead of reading VERCEL_OIDC_TOKEN env var directly, so the sandbox SDK works in production where the runtime sets the token via the x-vercel-oidc-token request header.


src/vercel/oidc/credentials.py
scotttrinh: Do we also want to improve this error message? Would be good to report which of these is set/unset and to make the message about providing project ID and team ID with the OIDC token more sensible.

scotttrinh (approved):

PR #108, opened 2026-04-21
Add security analysis via zizmor

scotttrinh:


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,24 @@
+name: GitHub Actions Security Analysis
```
scotttrinh, line 1: Since we added the test to our pytest executed cqa tests, do we still want this separate action? The nice thing about this action is that it is well integrated into the Advanced Security capabilities of GitHub Actions. It also runs even if the tests fail for other reasons, so feels like a nice-to-have.

fantix (approved):

.github/workflows/zizmor.yml
fantix: Sounds like a good addition to me! Maybe we can limit its run to only `.github/workflows/**` changes?

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,24 @@
...
+  zizmor:
+    name: Run zizmor
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write # Required for upload-sarif (used by zizmor-action) to upload SARIF files.
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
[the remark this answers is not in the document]
scotttrinh: Since this repo is public, we do not need to set explicit `contents: read` permissions.

.github/workflows/publish.yml
```diff
@@ -35,17 +37,26 @@ jobs:
...
-      - uses: astral-sh/setup-uv@v3
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: false
       - run: uv venv --python 3.14
       - name: Create tag from version
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           VERSION=$(python scripts/get-version.py)
```
elprans, line 61: This is where https://github.com/actions/github-script would probably be better.

```js
            await github.rest.git.createRef({
              owner: context.repo.owner,
              repo: context.repo.repo,
              ref: `refs/tags/${tag}`,
              sha,
            })
```

PR #109, opened 2026-04-21
workflow: multiple fixes

fantix:
* Added EntityConflictError to quietly retry
* Fixed local world bug in clearing hooks
* Fixed `workflow.sleep()` in local world
* Emulate EntityConflictError in local world with locks
* Sandbox: pass through importlib and basic asyncio to fix bugs:
  * Some third parties depend on importlib global state
  * CancelledError must not be recreated for TaskGroup to work
* Fixed out-of-order hook events
* Fixed hook disposal
* Fixed a bug in resume() when extra loop iteration is needed

### DIFF
--- a/src/vercel/_internal/workflow/worlds/local.py
+++ b/src/vercel/_internal/workflow/worlds/local.py
@@ -1,5 +1,6 @@
 import hashlib
 import json
+import math
 import os
 import pathlib
 import traceback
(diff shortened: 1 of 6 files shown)


src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -93,18 +96,23 @@ async def queue(
         *,
         deployment_id: str | None = None,
         idempotency_key: str | None = None,
+        delay_seconds: float | None = None,
```
scotttrinh, line 99: Why do we accept a `float` here if we `ceil` it to an `int` later? Not a blocker but I've also been trying to use `timedelta` for durations in Sandbox more thoroughly, and since this is an internal API, shouldn't be to hard to do that.


```suggestion
        delay: timedelta | None = None,
```

scotttrinh (approved):
No blockers. Approved since my only real comment is about internal API that can be refactored as needed.

src/vercel/_internal/workflow/worlds/local.py
fantix: Yeah, it's mirrored from the JavaScript code, but hitting this workers API mismatch (float vs int, used in HTTP header `Vqs-Delay-Seconds`). `timedelta` sounds like a good idea!

PR #118, opened 2026-05-14
Harden CI dependency installs

scotttrinh:
CI runs were inconsistent because normal PR checks resolved floating dependencies on each run. This change adds bounded direct dependency ranges, checks in `uv.lock`, and updates CI to install with `uv sync --locked` so PR and push builds use a reproducible dependency set. We still want visibility into upstream dependency breakages, so this also adds a weekly/manual latest-dependencies workflow that runs `uv sync --upgrade` within the allowed ranges and alerts us when dependency drift breaks the build.


pyproject.toml
```diff
@@ -28,22 +28,23 @@ sandbox = "vercel.sandbox.__main__:main"
...
+    "mypy>=1.20.2,<2",
+    "build<2",
+    "twine<7",
+    "zizmor>=1.24.1,<2",
 ]
 
 [tool.uv]
 default-groups = ["dev"]
 package = true
+exclude-newer = "1 week"
```
elprans: This is IMO too wide.  1 day should be sufficient as most supply chain attacks (at least for popular packages) are detected within hours if not minutes.

.github/workflows/latest-deps.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: false
+
+      - name: Sync dependencies
+        run: uv sync --upgrade --python ${{ matrix.python-version }}
```
elprans, line 38: What is the purpose of this?  Detecting breakage in minor versions of peer dependencies?

scotttrinh: Yes.

pyproject.toml
scotttrinh: Looks like we're standardizing on 2 days in JS land so I'll update it here.

cramforce (approved):

PR #120, opened 2026-05-19
allow to set async cache into context

nsidnev:
Split the `cache` key in the context object into `cache` and `async_cache` to allow both of them to be configured in the `set_context` and to retrieve the required cache instance in the `get_cache` method


src/vercel/cache/context.py
```diff
@@ -10,6 +10,7 @@
     "vercel_wait_until", default=None
 )
 _cv_cache: ContextVar[object | None] = ContextVar("vercel_cache", default=None)
+_cv_async_cache: ContextVar[object | None] = ContextVar("vercel_async_cache", default=None)
```
scotttrinh, line 13: This just means you can have both now, is that right?

scotttrinh (approved):

PR #128, opened 2026-05-28
Add `unstable.sandbox`

scotttrinh:
Introduces a new `unstable` package at `vercel.unstable` (and internals at `vercel._internal.unstable`) that currently exposes the latest v2 persistent Sandbox.

Looks like this (excerpt from included example):

```py
async def review_code(files: list[tuple[str, str]], review_agent: str) -> str:
    async with sandbox.create_sandbox(
        runtime="python3.13",
        execution_time_limit=timedelta(minutes=1),
    ) as box:
        await box.fs.mkdir("workspace")
        async with box.fs.batch() as batch:
            for path, content in files:
                batch.write_text(path, content)
            batch.write_text("workspace/review_agent.py", review_agent)

        await box.run_process(
            "python",
            ["workspace/review_agent.py", "workspace"],
            kill_after=timedelta(seconds=30),
            check=True,
        )
        return await box.fs.read_text("workspace/review.md")
```


src/vercel/_internal/unstable/sandbox/handles.py
```diff
@@ -0,0 +1,1156 @@
...
+        return self
+
+    async def update_network_policy(self, network_policy: JSONValue) -> "SandboxRuntimeSession":
+        payload = await self._client.update_runtime_session_network_policy_payload(
+            session_id=self.id, network_policy=network_policy
+        )
+        self._apply_payload(payload)
+        return self
+
+    async def mkdir(self, path: str, *, cwd: str | None = None, recursive: bool = True) -> None:
```
elprans: I wonder if we want a more robust Filesystem-like interface here.  We encapsulate snapshots and commands into sub-interfaces, I think we should do the same for fs.

src/vercel/_internal/unstable/sandbox/api_client.py
```diff
@@ -0,0 +1,1289 @@
...
+        snapshot_retention: SnapshotRetention | None = None,
+    ) -> SandboxState:
+        credentials = await self._credentials_factory()
+        request = _CreateSandboxRequest(
+            project_id=project_id or credentials.project_id,
+            name=name,
+            runtime=runtime,
+            source=source,
+            ports=ports,
+            timeout=parse_duration(execution_time_limit, MILLISECOND),
```
fantix: I remember the unit is "seconds" instead of "milliseconds" when time is `int`/`float`? I might have missed something here.

src/vercel/_internal/unstable/sandbox/handles.py
scotttrinh: Yeah, I like that idea. Honestly having `mkdir` as a separate method is already overkill considering we have the command interface. We could either expose a full `fs` module or remove it.

src/vercel/_internal/unstable/session.py
```diff
@@ -3,18 +3,26 @@
...
+from vercel._internal.polyfills import Self
 from vercel._internal.unstable.errors import VercelSessionClosedError, VercelSessionError
 from vercel._internal.unstable.options import ServiceOptions, merge_service_options
-
-if TYPE_CHECKING:
-    from vercel._internal.http import AsyncTransport, SyncTransport
-    from vercel._internal.unstable.sandbox.client import AsyncSandboxClient, SyncSandboxClient
+from vercel._internal.unstable.sandbox.api_client import SandboxApiClient
+from vercel._internal.unstable.sandbox.options import SandboxServiceOptions
+from vercel._internal.unstable.sandbox.service import SandboxService
```
fantix: Why did you replace the lazy-import with this global import? I feel the old lazy-import makes sense?

src/vercel/_internal/unstable/sandbox/runtime_common.py
```diff
@@ -0,0 +1,324 @@
...
+    return SandboxCommandLog(stream=stream, data=data)
+
+
+def _select_output(
+    snapshots: Sequence[_LogSnapshot], stream: Literal["stdout", "stderr", "both"]
+) -> str:
+    return "".join(data for source, data in snapshots if stream == "both" or source == stream)
+
+
+@dataclass(slots=True, eq=False, repr=False)
```
fantix: I found the `@dataclass`-es in this file are not quite necessary and misleading.

src/vercel/_internal/unstable/sandbox/sync_runtime.py
```diff
@@ -0,0 +1,824 @@
...
+            service=service,
+            session_id=lambda: self.current_session_id,
+            write_files_cwd=self._write_files_cwd,
+        )
+
+    @property
+    def current_session(self) -> SyncSandboxRuntimeSession | None:
+        return self._current_session
+
+    def _apply_payload(self, payload: SandboxState) -> None:
```
fantix: Worth extracting `_apply_payload()` into base class?

src/vercel/_internal/unstable/sandbox/api_client.py
```diff
@@ -0,0 +1,1304 @@
...
+            token=credentials.token,
+            params=query,
+            body=body,
+            headers=request_headers,
+        )
+
+        try:
+            response.read()
+        except RuntimeError:
+            await response.aread()
```
fantix: Maybe this can be pushed down to `transport.send(read_response="always")` (`always`/`non_success_only`/`never`), so that we can drop the `try...except`?

fantix (approved):
This looks great to me!

src/vercel/_internal/unstable/sandbox/api_client.py
scotttrinh: Ahh, good catch. We have to send it to the backend as milliseconds, but we need to parse this at the incoming boundary as numeric seconds or timedelta. See, this is why I hate unitless time duration values! 😆

src/vercel/_internal/unstable/sandbox/api_client.py
scotttrinh: Oh, that's a great idea, will do!

src/vercel/_internal/unstable/sandbox/api_client.py
scotttrinh: OK, I've done a thorough audit of the incoming duration units and they were all assuming milliseconds. I've switched this to "numerics mean seconds" (which matches the asyncio.Timeout API that we're matching here). We also carry around `timedelta` everywhere _internally_ so that only the boundaries have a unit-less duration value.

src/vercel/_internal/unstable/session.py
scotttrinh: I am trying to figure out how the latest typing annotations structures work, so went a little too far in the direction of not requiring `from __future__ import annotations` vs manually quoting type names. I've switched back to manual quoting + lazy importing.

If you have any suggestions as to how to approach this with more rigor, I'd love to hear what approaches work best these days!

src/vercel/_internal/unstable/sandbox/api_client.py
scotttrinh: went with `StrEnum` since that seems to be the preference of the rest of the team versus `Literal`, but this looks much better, thanks for the suggestion!

src/vercel/_internal/unstable/session.py
fantix: I believe lazy imports and manual quoting should work just fine here!

Alternatively, I'm thinking maybe it would be easier if the service constructor (a.k.a. `def sandbox_service()`) actually resides inside the service code instead of the session code, consuming session features through (internal) APIs on the session object. I don't think this is a necessary change though.

fantix (approved):

src/vercel/_internal/unstable/session.py
scotttrinh: Ahh, yeah, I think I get what you're saying: `SdkSession` can be used by the service rather than the session owning the service itself, which would flip the imports around. That _does_ seem more scalable for adding more services, so lemme explore that. Should be doable now and set us up for the next set of services.

scotttrinh: @fantix ⚡ https://github.com/vercel/vercel-py/pull/128/commits/5b709b194f67fe87e36636699f257c8fb4c7b078

Yeah, I think I like this better! Is this what you had in mind?

fantix: Yes exactly!

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+
+    async def is_dir(self, path: RemotePath, *, cwd: RemotePath | None = None) -> bool:
+        return await self._service.is_dir(
+            session_id=self._session_id(),
+            path=_coerce_remote_path(path),
+            cwd=None if cwd is None else _coerce_remote_path(cwd),
+            collect_output=self._collect_output,
+        )
+
+    async def listdir(
```
1st1, line 400: list_dir to align with other methods? Or this is inspired by shutil?

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+    __slots__ = ("_service", "stderr", "stdout")
+
+    def __init__(self, *, payload: ProcessState, service: SandboxService) -> None:
+        super().__init__(payload)
+        self._service = service
+        self.stdout, self.stderr = _text_readers(
+            lambda: service.process_logs_response(session_id=self._session_id, process_id=self.id)
+        )
+
+    async def refresh(self) -> Self:
```
1st1, line 107: I think we should add docstrings to all public API methods? What's the purpose of `refresh`?

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+        command: str,
+        args: Sequence[str] | None = None,
+        *,
+        cwd: str | None = None,
+        env: Mapping[str, str] | None = None,
+        sudo: bool = False,
+        kill_after: float | timedelta | None = None,
+        check: bool = False,
+        stdout: TextIO | int | None = None,
+        stderr: TextIO | int | None = None,
```
1st1, line 538: This can be a problem -- sometimes process output isn't decodable, especially if caught between utf opcodes or something.

Does the underlying vercel sandbox API operate on text or bytes?

1st1: In general, subprocess APIs are typically better with BytesIO for this reason. Users basically can decode on their own (potentially setting the codec to ignore decoding/encoding errors for this reason)

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+        return [Process(payload=state, service=self._service) for state in states]
+
+    async def refresh(self, *, include_system_routes: bool | None = None) -> Self:
+        payload = await self._service.get_runtime_session(
+            session_id=self.id, include_system_routes=include_system_routes
+        )
+        self._apply_payload(payload)
+        return self
+
+    async def extend_execution_time_limit(self, duration: DurationInput) -> Self:
```
1st1, line 667: This one is weird, just stating for the record.

I assume we do want to let the sandbox manage its timeout via the API for.. reasons, right?

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+        payload = await self._service.extend_runtime_session_timeout(
+            session_id=self.id, duration=parse_required_duration_seconds(duration)
+        )
+        self._apply_payload(payload)
+        return self
+
+    async def update_network_policy(self, network_policy: JSONValue) -> Self:
+        payload = await self._service.update_runtime_session_network_policy(
```
1st1, line 680: What's JSONValue here? I'd rather work with well-defined types than dicts

src/vercel/_internal/unstable/sandbox/async_runtime.py
```diff
@@ -0,0 +1,952 @@
...
+        self, files: Sequence[_WriteFile], *, cwd: RemotePath | None = None
+    ) -> None:
+        await self._service.write_files(
+            session_id=self._session_id(),
+            files=files,
+            cwd=self._write_files_cwd(cwd),
+        )
+
+    def batch(self, *, cwd: RemotePath | None = None) -> "SandboxFilesystemBatch":
+        return SandboxFilesystemBatch(write_files=lambda files: self._write_files(files, cwd=cwd))
```
1st1, line 359: Maybe rename to `.batch_io()`?

1st1 (approved):
Few comments here and there, let's have a quick iteration and then merge

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: Yeah, this is [the name from `os`](https://docs.python.org/3/library/os.html#os.listdir) but I also don't love it.

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: Good point, there is too much info locked up in the readme that needs to move into docstrings. Will fix.

To the actual question: the purpose of `refresh` is to get the latest state from the server, typically for updating things like the `status` and the `current_session_id`.

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: The underlying backend API is text-only via NDJSON lines, so yeah, it can't be bytes unless we just re-encode it into bytes. We can talk to the backend team about having something more low-level if they expose bytes, but as-is, we have text.

This seems like it's worth taking a pause on while we figure out what we want to do here. We can expose a text and byte interface where bytes are actually just re-encoded text for now and then switch that to some kind of mode on the backend side that sends buffers directly as `number[]`?

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: Yeah, this is a pretty well-used API in a lot of integrations where it is basically like an active keep-alive mechanism. Set the max execution time to something reasonable, and if some event happens before than, extend it again.

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: ahh, this is a vestige of how I built out the subresources, I missed this one. It's actually a pretty rich type, so I'll add the proper class structure here.

src/vercel/_internal/unstable/sandbox/async_runtime.py
scotttrinh: For now it's write-only. The original API was `write_files(files: Sequence(WriteFile))` and this is just a different API over that same underlying function. @elprans can you weigh in here on this design since it's mostly from your feedback? Or maybe a simple rename to `batch_write`?

Maybe eventually we'll have a combined read-write batch endpoint, but I'm not sure if we had that, how we'd even expose the read results back given this design?

scotttrinh:
Wanted to make a note here that we've decided to remove the `logs()` command from the process class for now. It has some utility, but it feels like a misfit compared to the other ways of consuming the stdio streams, so we're going to wait until we have a better understanding of the use case and see if there is a more natural API for it.

scotttrinh:
One more bit of feedback to record here. It doesn't block merging this PR since we can fast follow, but blocks the promotion of this `unstable` to the main namespace:

### Follow-up from the sandbox/session API discussion with @elprans 

The backend model is:

- A sandbox is the durable identity, configuration, snapshots, and session history. Omitting `name` generates one.
- A session is one billable execution. Creation always starts an initial session, and a sandbox has at most one active session.
- A snapshot preserves filesystem state between sessions and is used to resume a stopped sandbox.

The current `Sandbox.session()` API exposes sessions too prominently and suggests users can create independent active sessions. We agreed to keep normal lifecycle management at the sandbox level instead:

- `get_sandbox()` only fetches state and never resumes.
- `resume_sandbox()` ensures the sandbox has an active session.
- Awaiting `create_sandbox()` or `resume_sandbox()` performs no automatic cleanup.
- Both can be used as context managers, which always stop the active session on exit.
- A `create_sandbox()` context additionally destroys the sandbox by default.
- `create_sandbox(..., destroy=False)` stops the session but preserves the sandbox, snapshots, and history.
- `Sandbox.stop()` explicitly stops the active session.
- Session objects remain available for history, billing, diagnostics, and advanced inspection, but are not required for normal workflows.

Example:

```python
# Stops the session and destroys the sandbox on exit.
async with sandbox.create_sandbox(runtime="python3.13") as box:
    ...

# Stops the session but preserves the sandbox on exit.
async with sandbox.create_sandbox(name="development", destroy=False) as box:
    ...

# Fetches state without resuming or cleanup.
box = await sandbox.get_sandbox(name="development")

# Ensures an active session and stops it on exit.
async with sandbox.resume_sandbox(name="development") as box:
    ...

# Ensures an active session without automatic cleanup.
box = await sandbox.resume_sandbox(name="development")
await box.stop()
```

Nested or concurrent `resume_sandbox()` scopes may share the same active session; either scope exiting can stop it. We will document this and correctly surface stopped-session errors rather than adding runtime ownership or lease tracking.

PR #131, opened 2026-06-16
workflow: recover stranded local runs on restart

msullivan:
A sleep persists a wait and schedules a delayed queue message to wake it. Under `vercel dev` that message lives in an in-memory queue, so restarting the dev server mid-run loses the wake-up and strands the run in `running`.

On dev worker startup, LocalWorld now re-enqueues every non-terminal run, retrying via RetryLoop until the dev queue is reachable; workflow_handler turns any elapsed wait into a wait_completed and continues. Replay is idempotent, so this should be fine?

### DIFF
--- a/src/vercel/_internal/workflow/worlds/local.py
+++ b/src/vercel/_internal/workflow/worlds/local.py
@@ -1,8 +1,11 @@
+import asyncio
 import hashlib
 import json
+import logging
 import math
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


fantix:
Yeah I think this is good; I'm working on implementing the local queue out of `vercel dev` but it would also be an in-memory queue, so this will be still needed.

PR #132, opened 2026-06-17
workflow: check whether the replay log actually matches step args

msullivan:
Add a check in resume() that when creating a step, the arguments match
with what is in the log, and raise a NondeterminismError then.

Even with the sandbox, there are ways to get nondeterminism. The
biggest one is set iteration, which remains a big footgun.

Also check for when there is a type mismatch for a given position.

I'm not totally sure if we actually want to do this, though? In
particular, was it deliberate that we weren't?


fantix:
> was it (not check for when there is a type mismatch for a given position) deliberate that we weren't?

Not really, we were relying on the workflow being deterministic. So I think your fix here makes a lot of sense!

fantix (approved):

PR #136, opened 2026-06-22
workflows: Fix queue handlers requeueing of failed steps

msullivan:
Currently, the queue handlers, when the underlying handler returns a
retry timeout duration, will try to model_validate it as a
WorkflowInvokePayload, which is incorrect, since it might be a
StepInvokePayload or a HealthCheckPayload.

I think that in practice this probably doesn't mess things up *too*
much other than creating spurious error log traffic, since the queue
message won't have been ACKed and so should still get redelivered.


fantix (approved):
Ahh yes that is correct! Thank you for the fix

PR #138, opened 2026-06-23
workflows: set idempotency_key for workflow reinvocations triggered by waits

msullivan:
This matches the behavior of the JS library.

It requires changing handler from just returning a float to indicate a
re-enqueue to returning a new QueueContinuation that contains a
delay_seconds and an optional idempotency_key.

Another option, if we want to avoid the new type, would be get rid of
this generic re-enqueueing path and have it done explicitly in the
handlers.


fantix (approved):
Looks great!

PR #139, opened 2026-06-23
workflows: fix TOCTOU in write_json and make writes atomic for good measure

msullivan:
We were doing an `exists()` check but then doing the write afteward,
but this meant there was a race condition window.

Possibly we should be using sqlite?

### DIFF
--- a/src/vercel/_internal/workflow/worlds/local.py
+++ b/src/vercel/_internal/workflow/worlds/local.py
@@ -1,8 +1,10 @@
+import contextlib
 import hashlib
 import json
 import math
 import os
 import pathlib
+import tempfile
 import traceback
 from datetime import datetime
 from typing import Any, TypeVar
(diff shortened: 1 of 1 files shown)


fantix (approved):
> Possibly we should be using sqlite?

This depends on the workflow JS [CLI](https://github.com/vercel/workflow/tree/main/packages/cli)/[UI](https://github.com/vercel/workflow/tree/main/packages/web) - this toolset also consumes the local fs for workflow observability.

msullivan:
> > Possibly we should be using sqlite?
> 
> This depends on the workflow JS [CLI](https://github.com/vercel/workflow/tree/main/packages/cli)/[UI](https://github.com/vercel/workflow/tree/main/packages/web) - this toolset also consumes the local fs for workflow observability.

Oh, I didn't realize we had a fixed format. OK.

PR #140, opened 2026-06-23
workflows: put a Lock around local events_create

msullivan:
There are some read-modify-writes in it, so slap a per-run lock around
it to protect those.

The JS library does more fine-grained locking here but I don't think
the extra complication of that would really help much.


fantix (approved):
This is fine; we should probably also fix the vqs-py handler not to use `to_thread()` arbitrarily upon async functions.

PR #141, opened 2026-06-23
workflows: switch to raising EntityConflictError in some places in local world

msullivan:
There were some cases that raised RuntimeError, but they are reachable
due to legitimate races I think (though other fixes I've made should
make them less likely?), so we should raise EntityConflictError.

This should match what JS does.


fantix (approved):
Ah yes, good call! `EntityConflictError` was added in a later stage so these were missed out.

PR #143, opened 2026-06-23
workflows: add some os things to sandbox whitelist, only restrict callables

msullivan:
shutil, pathlib and third-party rich were both failing to import due
to sandbox interactions with os.

I'm adding getenv, getcwd, and PathLike to the whitelist.

I've also made it so that we only restrict callables, since we were
restricting the set `os.supports_dir_fd` which was breaking some
stuff. (We could also have added it to the whitelist but this seems a
little more general, though we will want to further generalize it
later.)


fantix (approved):
LGTM!

PR #144, opened 2026-06-24
workflow: rework sandboxing of sys.modules

msullivan:
The handling of sys.modules for sandboxes was not concurrency safe,
which is a problem because every queue request is handled in a
separate thread.

Instead of trying to do these big manipulations of sys.modules, follow
what Temporal does and install a proxying sys.modules shim that allows
each context to have its own set of imported modules.

This allows workflows to be run in fresh module environments where we
can ensure that everything is intercepted by the sandbox and that
won't have any interference with or from steps, while allowing steps
to run without needing to interact with proxy modules.

### DIFF
--- a/src/vercel/_internal/workflow/py_sandbox.py
+++ b/src/vercel/_internal/workflow/py_sandbox.py
@@ -1,14 +1,17 @@
 from __future__ import annotations
 
+import builtins
 import contextvars
 import dataclasses
 import datetime as _dt
+import importlib
 import os
 import random
 import sys
+import threading
 import types
 import weakref
-from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Callable, Iterator, Mapping, MutableMapping, Sequence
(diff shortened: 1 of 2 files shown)


src/vercel/_internal/workflow/py_sandbox.py
```diff
@@ -686,100 +765,114 @@ def updatecache(filename: str, module_globals: Any = None) -> list[str]:
...
-            _sandbox_refcount -= 1
-            if sys.meta_path is not None:
-                sys.meta_path.remove(finder)
-            sys.modules.clear()
-            sys.modules.update(orig_modules)
+        table["builtins"] = _real_sys_modules["builtins"]
+    return table
 
 
+# TODO: we probably want to support some form of sandbox caching
```
fantix, line 853: Yes! In a later stage sure!

fantix (approved):
Looks great!

PR #145, opened 2026-06-24
add .claude to .gitignore

msullivan:


scotttrinh (approved):
Sure! You can also add this to your `.git/info/exclude`, but this is reasonable.

PR #146, opened 2026-06-24
workflow: fix some races in step_handler

msullivan:
Currently we look up the step, do a check for exceeding max_retries
and for if the step is already in a terminal state, and if it is we
queue it up again and return None.

*Then* we try creating step. There are two issues here, one small and
one worse, that can happen if the step gets invoked twice
concurrently.

1. A check-then-use race with the attempt count can lead to exceeding
   max attempts.
2. If one step completes and then crashes without queueing, and then
   the other will fail to create the event and currently returns that
   all is well, so the requeue never happens.

Rework all this to not actually get the step_run and instead drive
everything off the result of events_create.

Also replace the runtime-side retry_after handling with processing
world-side reports of the errors.

I think this clearly worth fixing but it is also all pretty
low-probability, I realize now. I think that for two steps to be
running concurrently, one of them needs to fail to extend its lease
for quite a long time while still running and to still be in the very
early stages of processing the message.

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -641,6 +641,11 @@ async def workflow_handler(
     return w.QueueContinuation(delay_seconds=delay, idempotency_key=key)
 
 
+def _step_name_from_queue(queue_name: str) -> str:
+    """The step name encoded in a step queue topic."""
+    return queue_name.removeprefix("__wkf_step_")
+
+
 async def step_handler(
     message: Any,
     *,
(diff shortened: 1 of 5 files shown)


src/vercel/_internal/workflow/runtime.py
```diff
@@ -652,19 +657,49 @@ async def step_handler(
...
+            step.name,
+            timeout_seconds,
+        )
         return w.QueueContinuation(delay_seconds=timeout_seconds)
+    except w.EntityConflictError:
+        # Step already in a terminal state — a duplicate delivery, or a concurrent
+        # worker finished it. Re-enqueue the workflow so it observes the outcome,
+        # then ack. This is also the crash-recovery path for a step that finished
+        # but whose handler died before re-invoking the workflow.
+        logger.debug(f"Tried starting step {req.step_id!r}, but it has already finished")
```
fantix:
```suggestion
        logger.debug(f"Tried starting step %r, but it has already finished", req.step_id)
```

fantix (approved):
LGTM!

PR #147, opened 2026-06-26
Add `image` for Sandbox creation

scotttrinh:
Allow setting a VCR image name and optional tag/commit as the source of a new Sandbox.

See https://github.com/vercel/sandbox/pull/232


src/vercel/_internal/sandbox/models.py
```diff
@@ -467,6 +468,30 @@ class CreateSandboxRequest(_CreateModel):
...
+            raise SandboxValidationError(
+                [
+                    SandboxValidationIssue(
+                        path="runtime",
+                        message="runtime and image are mutually exclusive",
+                    )
+                ]
+            )
+        if isinstance(self.source, SnapshotSource) and (
+            self.runtime is not None or self.image is not None
```
marc-vercel: Nit: we support `snapshot` + `image` for now due to an edge case related to `WORKDIR` (I want to also remove it as both of them should be mutually exclusive), but for now we can't.

scotttrinh: @marc-vercel ⚡ 71e8fca

marc-vercel (approved):
LGTM, thanks!

PR #149, opened 2026-06-26
Convert to workspace, split `vercel.headers` and `vercel.oidc`

elprans:
Continuing with the monolithic `vercel` SDK package poses challenges
downstream from the volume and number of dependencies, especially where
we need to inject runtime code that relies on SDK facilities.

Remedy this by transitioning the monorepo to a uv workspace that
publishes multiple packages.  As a proof of concept, split out
`vercel.oidc` into standalone `vercel-oidc`.  There is no compatibility
breakage here as `vercel-oidc` is now a dependency of `vercel` and
contributes `vercel.oidc` as before.  Because `vercel.oidc` depends on
`vercel.headers`, split that out as well as `vercel-headers`.

While this isn't supposed to be breaking, bump version to 0.6.0 as
it's still a fairly invasive change.


scripts/no_root_build.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+MESSAGE = (
+    "This repository root is a uv workspace, not a publishable package. "
+    "Use `uv build --all-packages --no-sources` to build all packages, "
```
msullivan, line 5: Why --no-sources?

msullivan (approved):

scripts/no_root_build.py
elprans: Because we have `{ workspace = true }` deps now and [`uv` documents](https://docs.astral.sh/uv/guides/package/#building-your-package) that you're supposed to use `--no-sources` when publishing such packages.

PR #150, opened 2026-06-26
Switch to httpx2

elprans:
The original `httpx` library is no longer maintained, switch to
`httpx2`, which is a continuation of `httpx` and is maintained by the
Pydantic team.


scotttrinh (approved):
TIL, thanks!

PR #151, opened 2026-06-29
Add support for `image=` to `unstable.sandbox`

scotttrinh:
Matching `unstable.sandbox` PR for #147


scotttrinh:
We need to move this onto the main `vercel-sandbox` space. Should be simple.

PR #152, opened 2026-06-29
Add an agent skill for the release process

scotttrinh:
Add an agent skill for publishing packages to PyPI, reflecting the new flow where releases are driven locally by pushing signed package-version tags rather than by merging a branch-triggered publish PR. The skill captures the expected tag naming, publish verification, GitHub release title and notes conventions.


elprans:
The release process changed significantly and is intended to be human-operated (via `poe release`), closing.

PR #153, opened 2026-06-29
workflow: improvements to `@wf.step` decorator

msullivan:
* Allow passing `max_retries`
* Call `functools.update_wrapper(self, func)` in constructor so that function introspection works


fantix (approved):
Great!

PR #154, opened 2026-06-29
workflow: fix racy recreation of hook

msullivan:
We need to support two simultaneous invocations of a workflow both
creating a hook concurrently.

When trying to create a hook from a token, if it has already been
claimed, check whether hookId matches. If it does, it's the same hook,
not just one with a conflicting token, so raise EntityConflictError
instead of marking it as a hook conflict.


fantix (approved):
Ahh yes that is correct!

PR #155, opened 2026-06-30
workflow: harmonize error handling with JS a bit

msullivan:
Add some error classes that JS didn't have, and catch one of the new ones (HookNotFoundError) when disposing a hook.

### DIFF
--- a/src/vercel/_internal/workflow/worlds/local.py
+++ b/src/vercel/_internal/workflow/worlds/local.py
@@ -277,7 +277,7 @@ class LocalWorld(w.World):
                 hook = read_json(hook_path, w.Hook)
                 if hook is not None and hook.token == token:
                     return hook
-        raise w.HookNotFoundError(token)
+        raise w.HookNotFoundError(token=token)
 
     async def events_create(self, run_id: str | None, data: w.Event) -> w.EventResult:
         # run_created has no existing entity to race on — its create is guarded by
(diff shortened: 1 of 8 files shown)


src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -277,7 +277,7 @@ async def hooks_get_by_token(self, token: str) -> w.Hook:
                 hook = read_json(hook_path, w.Hook)
                 if hook is not None and hook.token == token:
                     return hook
-        raise RuntimeError(f"Hook with token {token!r} not found")
+        raise w.HookNotFoundError(token)
```
fantix: maybe worth a note here that, `token` is passed to - namely - `correlation_id` and it shouldn't be used as a hook ID?

fantix (approved):
LGTM!

src/vercel/_internal/workflow/worlds/local.py
msullivan: I'm not certain I follow this comment? hook IDs aren't specified by users

fantix: Right, I mean the `HookNotFoundError` constructor is taking a `correlation_id` but we are feeding in a user token here.

msullivan: Oh good catch. I've made it take separate correlation_id and token kwargs now

PR #156, opened 2026-07-02
Rework resume() to better handle workflows with parallel tasks

msullivan:
Having concurrent tasks in workflows is subtle, but its workable
because async concurrency is not real concurrency and is in principle
at least possible to make deterministic.

The core problem I was running into was resume() deciding to suspend
because there was a suspension pending, even though it had woken up
some other task that had productive work to do.
(The particular disasterous case that prompted all this was something
along the lines of one task waiting on a hook while another task
needed to run a step to publish the label of the hook. If that step
didn't trigger, deadlock.)

In order to fix that problem, we need to only suspend if no
suspensions were activated by resume() and when there are no ready
tasks in the event loop (since those might launch more steps, etc).
Doing this while preserving the same ordering of task executions
on all replays is a little subtle. The key idea is that it needs
to behave as if the events in the log were applied one at a time.
(Since they may have come in originally one at a time!)

So what we do is:
 * Arrange for resume() to always and only be called when the
   event loop ready queue is empty
 * Only activate one suspension per execution, so that it can
   resolve without interference.
 * Only cancel when the event log is exhausted and we did no work

--

Since we inspect `_ready`, this requires that we do this in an
isolated event loop. That's already the case for our vercel queue
based environments, but to avoid relying on this I make the workflows
run in its own isolated loop directly.

Eventually we'll probably want to have our own custom event loop I
think.

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -192,6 +192,7 @@ class WorkflowOrchestratorContext:
             token = self._ctx.set(self)
             try:
                 self._fut = asyncio.ensure_future(obj.func(*args, **kwargs))
+                self.resume_wrapper()  # arm the resume callback
             finally:
                 self._ctx.reset(token)
             return await self._fut
@@ -200,8 +201,6 @@ class WorkflowOrchestratorContext:
         input_data = b"json" + json.dumps((args, kwargs), sort_keys=True).encode()
         sus = Suspension(correlation_id=f"step_{self.generate_ulid()}", step=step, input=input_data)
         self.suspensions[sus.correlation_id] = sus
-        if self.resume_handle is None and not self._suspended:
-            self.resume_handle = asyncio.get_running_loop().call_soon(self.resume)
         return await sus.future
 
     async def run_wait(self, param: int | float | datetime | str) -> None:
@@ -210,8 +209,6 @@ class WorkflowOrchestratorContext:
             resume_at=(parse_duration_to_date(param)),
         )
         self.suspensions[wait.correlation_id] = wait
-        if self.resume_handle is None and not self._suspended:
-            self.resume_handle = asyncio.get_running_loop().call_soon(self.resume)
         await wait.future
 
     def create_hook(self, token: str | None, hook_cls: type[T]) -> core.HookEvent[T]:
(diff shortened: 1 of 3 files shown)


msullivan:
https://github.com/vercel-labs/seal/pull/41 has a bunch of code that depends on this, including some targeted tests

src/vercel/_internal/workflow/runtime.py
```diff
@@ -189,19 +210,36 @@ async def run_workflow(self: Self, workflow_run: w.WorkflowRun) -> Any:
...
+                # allowed to perform any IO-full operations, and
+                # resume/resume_wrapper require that it run in an
+                # isolated loop.
+                #
+                # So hide our existing loop and run everything in a
+                # fresh loop. We explicitly specify a factory because
+                # we look at ._ready, which might not exist in uvloop
+                # etc.
+                # TODO: Use a custom loop implementation.
+                asyncio._set_running_loop(None)
```
fantix: This is not enough in Python 3.14, which started to [track current task by thread](https://github.com/python/cpython/issues/129898), and causes the following error:

```
Exception in callback <_asyncio.TaskStepMethWrapper object at 0x7f6fa53176d0>()
handle: <Handle <_asyncio.TaskStepMethWrapper object at 0x7f6fa53176d0>()>
Traceback (most recent call last):
  File "/var/lang/lib/python3.14/asyncio/events.py", line 94, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Cannot enter into task <Task pending name='Task-5'
    coro=<WorkflowOrchestratorContext.run_workflow.<locals>.inner() running at
        /var/task/_vendor/vercel/_internal/workflow/runtime.py:213>
    cb=[_run_until_complete_cb() at /var/lang/lib/python3.14/asyncio/base_events.py:181]>
  while another task <Task pending name='Task-4'
    coro=<_await_result() running at
        /var/task/_vendor/vercel/workers/_queue/subscribe.py:124>
    cb=[_run_until_complete_cb() at /var/lang/lib/python3.14/asyncio/base_events.py:181]>
  is being executed.
```

fantix (changes requested):
Other than Python 3.14 support, I think this PR makes sense!

fantix (approved):

PR #162, opened 2026-07-09
[workflows sdk] support namespacing workflows

ricardo-agz:
Adds support for namespacing workflows into prefixed topics. This is necessary to support multiple workflow entrypoints in a project, e.g. for a Services project with a Django service with workflows + a FastAPI service with workflows

Otherwise, both of these workflow services would race for messages on the same topic and would fail as they have different registries. 

e.g. 

```toml
[[tool.vercel.workflows]]
entrypoint = "flows_a:wf"

[[tool.vercel.workflows]]
entrypoint = "flows_b:wf"
```

Each of these workflows would need to be namespaced:
```python
# flows_a.py
wf = Workflows("a")
# topic prefix = __a_wkf_*

# flows_b.py
wf = Workflows("b")
# topic prefix = __b_wkf_*
```

This will also be necessary to support a project with both JS workflows + Python workflows.

**Note:** While the JS workflows sdk also supports namespacing, they support it via an environment variable. We could possibly consider doing this as well although the in-code definition is helpful in cases where a service can invoke a 2 separate workflows, e.g. FastAPI needs to be able to invoke both Workflow "a" and Workflow "b" and needs to know which topic to enqueue the message to

### DIFF
--- a/src/vercel/_internal/workflow/core.py
+++ b/src/vercel/_internal/workflow/core.py
@@ -12,6 +12,7 @@ import pydantic
 from vercel._internal.polyfills import Self
 
 from . import py_sandbox
+from .world import validate_queue_namespace
 
 if TYPE_CHECKING:
     from . import world as w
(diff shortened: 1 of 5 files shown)


src/vercel/_internal/workflow/core.py
```diff
@@ -141,7 +151,10 @@ async def resume(self, token_or_hook: str | w.Hook, **kwargs) -> w.Hook:
 
 
 class Workflows:
-    def __init__(self, *, as_vercel_job: bool = True):
+    def __init__(self, namespace: str | None = None, *, as_vercel_job: bool = True):
```
msullivan: How about making `namespace` keyword only?

msullivan (approved):

PR #171, opened 2026-07-10
workflow: block ambient random, add workflow.random()

msullivan:
random.random()/.randint()/etc previously worked transparently inside the sandbox by silently redirecting to a per-run seeded instance. Replace that with the now()/time_ns() pattern: block the ambient module surface entirely and require the explicit workflow.random() accessor, which returns a memoized per-run random.Random instance decoupled from the internal ID-generation PRNG.

This is a breaking change, though?

### DIFF
--- a/src/vercel/_internal/workflow/core.py
+++ b/src/vercel/_internal/workflow/core.py
@@ -4,6 +4,7 @@ import dataclasses
 import datetime
 import functools
 import json
+import random as _random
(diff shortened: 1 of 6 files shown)


fantix (approved):
Yeah, this is breaking esp. libraries using `random` internally, but I think that's okay - the semantic of such library was "non-deterministic" anyways and should not be used in workflow functions.

PR #186, opened 2026-07-20
Parallelize workspace Poe checks

elprans:
Make poe tasks run in parallel by default (opt out via
`WORKSPACE_POE_PARALLEL=0`).  Makes `pre-commit` check run in ~2s and
`pre-push` in ~10s.  This also enables `pytest-xdist`.

### DIFF
--- a/scripts/poe/README.md
+++ b/scripts/poe/README.md
@@ -45,15 +45,19 @@ verbosity = -1
 
 The shared include defines these Poe tasks:
 
-- `lint`: runs `$RUFF_CHECK`, then `$RUFF_FORMAT`.
+- `lint`: runs `$RUFF_CHECK` and `$RUFF_FORMAT` in parallel.
 - `fix`: runs `$RUFF_CHECK_FIX`, then `$RUFF_FORMAT_FIX`.
-- `typecheck`: runs `$POE typecheck-mypy`, then `$POE typecheck-ty`.
+- `typecheck`: runs `$POE typecheck-mypy` and `$POE typecheck-ty` in parallel.
 - `typecheck-mypy`: runs `$MYPY`.
 - `typecheck-ty`: runs `$TY`.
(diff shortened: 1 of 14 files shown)


scripts/poe/README.md
```diff
@@ -45,16 +45,20 @@ verbosity = -1
 
 The shared include defines these Poe tasks:
 
-- `lint`: runs `$RUFF_CHECK`, then `$RUFF_FORMAT`.
+- `lint`: runs `$RUFF_CHECK` and `$RUFF_FORMAT` in parallel.
 - `fix`: runs `$RUFF_CHECK_FIX`, then `$RUFF_FORMAT_FIX`.
```
scotttrinh, line 49: I know we didn't change this here, but doesn't ruff fix already run the check?

scripts/poe/README.md
```diff
@@ -45,16 +45,20 @@ verbosity = -1
 
 The shared include defines these Poe tasks:
```
scotttrinh, line 46: I can follow-up myself on this, but seems like a good DX improvement here would be to add documentation to the poe commands themselves in addition (or instead of?) the manual documentation in the readme and AGENTS.md?

scotttrinh (approved):
👏 🏆

scripts/poe/README.md
elprans: `ruff fix` also fixes/formats, which is not the same as read-only checking.

scripts/poe/README.md
elprans: That's already done (or maybe I misunderstand what you mean?).  See CONTRIBUTING.md

scotttrinh: Oh sorry, I meant to include this link for what I meant: https://poethepoet.natn.io/guides/help_guide.html

PR #192, opened 2026-07-21
Use lograil to provide interactive progress for parallel poe tasks

elprans:
Pre-commit and pre-push hooks now show nice spinners and progressbars
and most QA-related poe tasks as well.


scripts/githooks/pre-commit.checks.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env sh
 set -eu
 
-. "$(dirname "$0")/helpers/run-quietly.sh"
-
-run_quietly "pre-commit checks" uv run poe pre-commit
+printf '\n'
+if FORCE_COLOR=1 CLICOLOR_FORCE=1 PY_COLORS=1 uv run poe pre-commit; then
```
scotttrinh, line 5: Will this interact in a weird way with agents when they make a commit?

scotttrinh (approved):
Looks good to me, one non-blocking question.

scripts/githooks/pre-commit.checks.sh
elprans: Don't let your agents commit :-). But also, proper harnesses run their shells in a PTY, so everything should work.

PR #194, opened 2026-07-22
Move vercel tests under package

elprans:
Tests for the monolithic SDK part are moved to src/vercel/tests instead
of being at the repo root which makes poe scope more reasonable.


scotttrinh (approved):
Change looks clean!

PR #197, opened 2026-07-27
Change credentials_factory

scotttrinh:
For sync usage, we should take a normal sync function instead of requiring that you give us an async shaped, non-suspending function that we pass through iter_coroutine.

Also export the required types here so consumers can properly type their usage.

### DIFF
--- a/src/vercel-sandbox/vercel/sandbox/sync.py
+++ b/src/vercel-sandbox/vercel/sandbox/sync.py
@@ -47,8 +47,12 @@ from vercel.sandbox._internal.models import (
     TagFilter,
     TarballSource,
 )
-from vercel.sandbox._internal.options import SandboxServiceOptions
-from vercel.sandbox._internal.service import SandboxService, get_sandbox_service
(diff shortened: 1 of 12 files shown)


src/vercel-sandbox/vercel/sandbox/sync.py
```diff
@@ -47,8 +47,12 @@
     TagFilter,
     TarballSource,
 )
-from vercel.sandbox._internal.options import SandboxServiceOptions
-from vercel.sandbox._internal.service import SandboxService, get_sandbox_service
+from vercel.sandbox._internal.options import (
+    SandboxCredentials,
+    SyncSandboxCredentialsFactory,
+    SyncSandboxServiceOptions as SandboxServiceOptions,
```
scotttrinh, line 53: Should we just keep the `Sync` prefix here? We have a mix of unprefixed and prefixed stuff in the `vercel.sandbox.sync` package, probably makes sense to be consistent here. Thoughts?

1st1 (approved):

PR #205, opened 2026-07-29
Vercel Connect SDK

kajogo777:
Depends on #206 and #207 

## Implement Vercel Connect functions:
- `get_token`
- `get_token_response`
- `revoke_token`
- `start_authorization`
- `get_connector_metadata`
- `verify_connect_webhook`

## Out of scope experimental TS SDK features:
- `experimental.start_installation`

## Changes to `vercel-oidc`:
`verify_connect_webhook` needed to verify Vercel's OIDC token, so I added `verify_vercel_oidc_token` and `verify_vercel_oidc_token_async` behind optional dep `vercel-oidc[verify]` because it adds `pyjwt` and `crypto`. I was not sure why verifying wasn't there in the first place

## Difference in cache implementation of TS SDK:
- cache is keyed by vercel token + content
- make field and array order irrelevant when caching tokens

## AI workflow disclosure:
- created some sample scripts with TS SDK to understand how Vercel Connect works, used Slack integration for testing -> generated a report explaining what Vercel Connect is and what it does
- analyzed vercel-py conventions to generate a report
- analyzed TS SDK implementation to generate a report
- compiled a python SDK implementation plan in a report
- make design choices on scope (excluding experimental features), deps to use (where to use pydantic, which jwt lib to use), missing OIDC changes upstream, and token caching implementation
- code gen
- multiple code review agents GPT 5.6 sol + Claude opus 5 (4 turns)
- created some sample scripts with Python SDK to test from consumer perspective


src/vercel-connect/vercel/connect/_internal/cache.py
```diff
@@ -0,0 +1,311 @@
...
+                    _canonical(serialize_authorization_detail(detail)),
+                    sort_keys=True,
+                    separators=(",", ":"),
+                )
+                for detail in authorization_details
+            )
+        ),
+        # Hash, never store: two platform identities must never share an entry,
+        # and the raw token must not be retrievable from the cache.
+        "identity": hashlib.sha256(vercel_token.encode()).hexdigest(),
```
dvoytenko: Token rotates without changing its identity. This is not super harmful though, so if this is a real case in this API, I'm ok keeping it. If you want to get very precise though, you can do this:

1. Try to parse as JWT
2. Extract iss, aud, and sub claims. Concatenate them. Use as the cache key
3. If (1) or (2) fail - use sha256 on the whole token.

src/vercel-oidc/vercel/oidc/verify.py
```diff
@@ -0,0 +1,546 @@
...
+
+
+def verify_vercel_oidc_token(
+    token: str,
+    *,
+    project_id: str | None = None,
+    environment: str | None = None,
+    owner_id: str | None = None,
+    audience: str | Sequence[str] | None = None,
+    issuer: str = VERCEL_OIDC_ISSUER,
```
dvoytenko: Is this customizable for tests? We typically very strict against making it customizable in JS SDKs, but maybe this is ok for Python.

src/vercel-oidc/vercel/oidc/verify.py
```diff
@@ -0,0 +1,546 @@
...
+    token_project = claims.get("project_id")
+    token_environment = claims.get("environment")
+    token_owner = claims.get("owner_id")
+
+    if token_environment != environment:
+        raise VercelOidcVerificationError(
+            f"token environment {token_environment!r} does not match expected {environment!r}"
+        )
+
+    if project_id == _WILDCARD or token_project == _WILDCARD:
```
dvoytenko: Might be slightly misleading, but I'm not aware of a case where we'd have project_id claim be "*".

src/vercel-connect/vercel/connect/_internal/models.py
```diff
@@ -0,0 +1,213 @@
...
+class ConnectGitHubAppInstallationAuthorizationDetail:
+    """A GitHub App installation authorization detail (RFC 9396)."""
+
+    org: str | None = None
+    permissions: Sequence[str] | None = None
+    repositories: Sequence[str] | None = None
+    type: str = "github_app_installation"
+
+    def __post_init__(self) -> None:
+        _validate_string_sequence("permissions", self.permissions)
```
dvoytenko: I think TS version (and API) accept `string | string[]` for these types. But it's ok to keep it narrower here if more canonical.

dvoytenko (changes requested):

src/vercel-connect/vercel/connect/_internal/api_client.py
```diff
@@ -0,0 +1,436 @@
...
+except PackageNotFoundError:  # pragma: no cover - bundled distribution
+    try:
+        VERSION = _pkg_version("vercel-connect-bundle")
+    except PackageNotFoundError:
+        VERSION = "development"
+
+PLATFORM = platform.uname()
+USER_AGENT = (
+    f"vercel-connect/{VERSION} (Python/{sys.version}; {PLATFORM.system}/{PLATFORM.machine})"
+)
```
scotttrinh, line 59: Seeing this again makes me feel like we should centralize this functionality into `vercel-internal-core`, maybe in the transport, to avoid needing to redefine this everywhere and keep them up to date.

No action needed in your PR, but leaving a note here, and we'll (have an agent) circle back on this.

src/vercel-connect/vercel/connect/_internal/api_client.py
```diff
@@ -0,0 +1,436 @@
...
+# Server error codes are machine tokens such as `user_authorization_required`.
+# A prose sentence in the same field is a message, not a code.
+_ERROR_CODE_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.:-]{0,63}$")
+
+_ERROR_CLASSES: Mapping[str, type[ConnectApiError]] = {
+    "no_token": NoValidTokenError,
+    "user_authorization_required": UserAuthorizationRequiredError,
+    "client_installation_required": ConnectorInstallationRequiredError,
+    "connector_installation_required": ConnectorInstallationRequiredError,
+}
```
scotttrinh: Since we already want/need to define the codes we care about, why have the regexp? Looking at the places we use `_ERROR_CODE_PATTERN` I feel like it would be marginally safer to check it against the _ERROR_CLASSES key set.

src/vercel-connect/vercel/connect/_internal/api_client.py
```diff
@@ -0,0 +1,436 @@
...
+            client_url=self.client_url,
+            created_at=None if self.created_at is None else _from_epoch_ms(self.created_at),
+            updated_at=None if self.updated_at is None else _from_epoch_ms(self.updated_at),
+            vendor=dict(self.data or {}),
+            extra=extra,
+        )
+
+
+def _from_epoch_ms(value: int) -> datetime:
+    return datetime.fromtimestamp(value / 1000, tz=timezone.utc)
```
scotttrinh: Small ask: can we move this to `vercel._internal.core.time`? Feels generally useful. I know that module is currently very duration oriented, but it's meant to grow into a general datetime utility module.

src/vercel-connect/vercel/connect/_internal/api_client.py
```diff
@@ -0,0 +1,436 @@
...
+        if installation_id is not None:
+            body["installationId"] = installation_id
+        if audience is not None:
+            body["audience"] = list(audience)
+        if resources is not None:
+            body["resources"] = list(resources)
+        if authorization_details is not None:
+            body["authorizationDetails"] = [
+                _serialize_authorization_detail(detail) for detail in authorization_details
+            ]
```
scotttrinh: Not a blocker, per se, but this is a bit inconsistent with how I structured the sandbox API client. In sandbox, I made the request body also be a typed model since we're often doing transformations within the function body here. That avoids accidentally losing the typed boundary we have in the function call to an loose `dict[str, any]`.

Ditto for the other request bodies in the other methods.

src/vercel-connect/vercel/connect/_internal/cache.py
kajogo777: Done, except the identity comes from verified claims

src/vercel-connect/vercel/connect/_internal/models.py
```diff
@@ -0,0 +1,213 @@
+"""Public value types for the Connect SDK surface."""
```
scotttrinh: We discussed just centralizing on Pydantic models for everything, per @1st1 , so let's go ahead and start doing that here. I'll open a separate PR to do this for Sandbox as well.

src/vercel-oidc/vercel/oidc/verify.py
kajogo777: Removed, and it surfaced a bug: the service publishes both `https://oidc.vercel.com` and `https://oidc.vercel.com/<team>` signed by one global key, so exact-match `iss` was rejecting every team-scoped token

src/vercel-oidc/vercel/oidc/verify.py
kajogo777: Removed

src/vercel-connect/vercel/connect/_internal/models.py
kajogo777: Made it narrower, because str turned out to satisfy Sequence[str]

src/vercel-connect/vercel/connect/_internal/service.py
```diff
@@ -0,0 +1,468 @@
...
+
+def is_detached_interactive_auth() -> bool:
+    """Whether device-code authorization is the configured default."""
+    return os.environ.get(_DETACHED_ENV, "").strip().lower() == "detached"
+
+
+def _from_epoch(value: Any) -> datetime | None:
+    if not isinstance(value, (int, float)):
+        return None
+    return datetime.fromtimestamp(value, tz=timezone.utc)
```
scotttrinh: Same note here about moving to the time module.

dvoytenko (approved):

src/vercel-connect/README.md
```diff
@@ -0,0 +1,139 @@
...
+exact `match`/`case` narrowing, and immutability, so a subject cannot be mutated
+after a credential has been cached against it. The cost is the empty `()` on the
+app subject, which is the common case.
+
+## Examples
+
+Runnable scripts for each surface live in [`examples/`](examples/), covering token
+minting, the consent flow, cache behaviour, connector introspection, the sync
+surface, and webhook verification. Start with
+[`connect_01_app_token.py`](examples/connect_01_app_token.py).
```
scotttrinh: We removed these, right? I still think we should have runnable examples in our repos (and run them in CI), but we can add them back later if we get broad support for that from the team.

src/vercel-connect/README.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+Verification pins the issuer to `https://oidc.vercel.com`, allows only RS256, and
+fails closed when the expected project and environment cannot be resolved. It
+accepts any valid Vercel OIDC token for this project and environment; it is not
+pinned to a specific connector or deployment.
+
+## Configuration
+
+`ConnectServiceOptions` is the configuration and test seam:
```
scotttrinh: Maybe this reads a bit better:


```suggestion
To configure advanced options, use a `session` context manager, and pass `ConnectServiceOptions`.
```

src/vercel-connect/README.md
```diff
@@ -0,0 +1,139 @@
...
+```sh
+vercel link
+vercel env pull    # writes VERCEL_OIDC_TOKEN into .env.local
+```
+
+The connector must be attached to your project and enabled for the target
+environment, or every call fails.
+
+Use a plain `with` block and `vercel.connect.sync` together; mixing an async call
+into a sync session (or the reverse) is rejected.
```
scotttrinh: This warning seems out of place here.

src/vercel-oidc/vercel/oidc/verify.py
scotttrinh: It shouldn't be _required_ for testing since in a test case we'd control the contents of the token, right? I'm fine with dropping this as an argument and hardcoding `VERCEL_OIDC_ISSUER` into the verification, at least in this public signature.

src/vercel-oidc/vercel/oidc/verify.py
```diff
@@ -0,0 +1,546 @@
...
+    with _jwks_lock:
+        _jwks_fetch_in_progress.discard(url)
+
+
+def _fetch_is_in_progress(url: str) -> bool:
+    with _jwks_lock:
+        return url in _jwks_fetch_in_progress
+
+
+def _select_and_record(url: str, kid: str, document: dict[str, Any]) -> Any:
```
scotttrinh: We have a lot of `Any` floating around in this module. I _think_ this might be due to the typing coming from `pyjwt`, but can we do some checking and casting at the boundaries to reduce the number of `Any` values that are actually passed around within our functions?

src/vercel-connect/vercel/connect/_internal/state.py
```diff
@@ -1,13 +1,99 @@
...
+DEFAULT_SCOPES_SENTINEL = "*"
+_LOCAL_HOSTS = ("localhost", "127.0.0.1", "[::1]", "::1")
+
+
+def _is_local_host(host: str) -> bool:
+    host = host.lower()
+    return host in _LOCAL_HOSTS or host.endswith(".localhost")
+
+
+class _ConnectorRequest(ConnectModel):
```
scotttrinh: This is a step in the right direction! The last step is to move the "serialize to API wire shape" functionality into the model as well to avoid having to build the loose `dict[str, Any]` later. See https://github.com/vercel/vercel-py/blob/b8dfa939c19d0e7a6eeb145292bddc798a3f03ca/src/vercel-sandbox/vercel/sandbox/_internal/api_client.py#L119-L129 for how we did it in Sandbox.

src/vercel-oidc/README.md
```diff
@@ -26,3 +26,43 @@ VERCEL_OIDC_TOKEN=$(vc project token some-project) some-command
...
+from vercel.oidc import verify_vercel_oidc_token_identity
+
+identity = verify_vercel_oidc_token_identity(token)  # stable across a refresh
+```
+
+The signature, issuer and expiry are verified, and `iss`, `aud` and `sub` are
+reduced to an opaque digest that carries no credential and is safe to log. This
+is **not** an authorization check — it returns no claims and does not check the
+project, environment, owner or audience. Use `verify_vercel_oidc_token()` for
+that.
```
scotttrinh: I get the utility of this function for our own usage in the cache, but I wonder if we should just have `verify_vercel_oidc_token` return `(identity, claims)` so we don't have this "verify" that doesn't full verify method. Would that work for our usage?

src/vercel-connect/README.md
1st1: If both of you feel strongly about having them - let's have them, just make sure you have automation to keep them in running order

src/vercel-oidc/README.md
kajogo777: we can't really verify here to enable connect to work locally and remotely inside projects, the reason is that `VERCEL_PROJECT_ID` / `VERCEL_TARGET_ENV`|`VERCEL_ENV` are required for verification, and I believe they'll be rarely set in local dev envs

happy to rename it though to something other that `verify_*`

src/vercel-connect/vercel/connect/_internal/api_client.py
kajogo777: are we fine with not being able to recognize new / unknown error codes?

scotttrinh: I'm personally fine with that, but couldn't we just do the same thing we're doing now? What error do you get if there is an unknown code in the current implementation?

src/vercel-oidc/README.md
scotttrinh: Yeah, that makes sense, and I had a suspicion that the reason this function exists is for this kind of edge case. For `VERCEL_PROJECT_ID` we could always fall back to the project ID from the OIDC token we have locally, but for the env, I'm not sure we could/should hardcode a "fallback" to development or something like that.

What about `extract_`?

src/vercel-connect/README.md
scotttrinh: Yeah, we already have automation that runs the examples on CI, and I run them extensively when doing local development.

src/vercel-connect/vercel/connect/_internal/api_client.py
kajogo777: agreed, left it alone for now

src/vercel-connect/vercel/connect/_internal/api_client.py
kajogo777: added more missing errors, and now the rest won't be addressed

src/vercel-connect/vercel/connect/_internal/api_client.py
kajogo777: moved it

src/vercel-connect/vercel/connect/_internal/api_client.py
kajogo777: done

src/vercel-connect/vercel/connect/_internal/state.py
kajogo777: done

src/vercel-oidc/vercel/oidc/verify.py
kajogo777: agreed, fixed this

src/vercel-oidc/README.md
kajogo777: made it `resolve_vercel_oidc_token_identity` to avoid confusion

src/vercel-connect/README.md
kajogo777: removed them for now

src/vercel-connect/README.md
kajogo777: moved it

changes/vercel-connect/error-taxonomy.feature.md
```diff
@@ -0,0 +1,13 @@
+Add error classes for the codes Connect forwards from an upstream provider's
```
scotttrinh: We can remove/consolidate the multiple changelog fragments into a single one. I think the original connect-sdk.feature.md is comprehensive enough without this additional detail, so I'd opt for remove over consolidate, but your call if you want to make it a bit more detailed.

scotttrinh (changes requested):
Looks good, thanks! One more bit of clean up.

scotttrinh (approved):
🏆  Thanks for the hard work here!

PR #207, opened 2026-07-30
Bind async session clients to the running event loop

kajogo777:
## Problem

The default async `SdkSession` is a process-wide singleton that held one
`httpx.AsyncClient` for the life of the process. That client pools keep-alive
sockets owned by the event loop that opened them, so a second loop reusing the
session sends the request, reads the response, and then raises
`RuntimeError: Event loop is closed` while closing the connection.

The caller sees an error for work the server already performed, so a retry can
duplicate a non-idempotent operation. From the new test's server, before the fix:

```
client: ok=3 failed=3   server: requests_processed=6
```

It also alternates rather than self-healing: the failed attempt discards the
stale socket, the next call opens a fresh one, and that one goes stale when its
loop closes.

Who hits it: any async API used without an explicit `session()` across more than
one loop. Scripts and CLIs calling `asyncio.run` twice, test suites with a loop
per test, and `vercel.sandbox`, which resolves its service from the default
session. The same root cause was previously worked around in `vercel-cache` by
opening a client per call (#21, #23).

## Fix

Keep one client per event loop on the session, and let the transport resolve it
per request. A check inside `get_transport()` would not be enough: services
capture the transport once at construction and are memoized on the session, so
`get_transport()` is never called again and the stale client is reached through
the cached service. Resolving per request also makes a session shared by threads
that each run their own loop safe, which swapping a single client would not.

Clients for closed loops are dropped when a new one is created, because `aclose`
has to run on the loop that owns the sockets.

## Tests

Three regression tests, all verified failing before the fix and passing after.
They use a loopback keep-alive server rather than `respx`, since a mocked
transport never pools a socket and cannot reproduce this.

1. repeated `asyncio.run` through the default session, asserting no client errors
   and one server-side request per call
2. a memoized service holding the transport across loops, which is what makes the
   `get_transport()`-only fix insufficient
3. four threads with concurrent loops sharing one session, each resolving its own
   client

### DIFF
--- a/src/vercel-internal-core/vercel/_internal/core/http/transport.py
+++ b/src/vercel-internal-core/vercel/_internal/core/http/transport.py
@@ -6,7 +6,7 @@ import abc
 import json
 import queue
 import threading
-from collections.abc import AsyncIterator, Iterator, Mapping, Sequence
+from collections.abc import AsyncIterator, Callable, Iterator, Mapping, Sequence
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from dataclasses import dataclass
 from datetime import timedelta
@@ -713,8 +713,24 @@ class SyncTransport(BaseTransport):
 class AsyncTransport(BaseTransport):
     _client: httpx.AsyncClient
 
-    def __init__(self, client: httpx.AsyncClient) -> None:
+    def __init__(self, client: httpx.AsyncClient | Callable[[], httpx.AsyncClient]) -> None:
+        """Take a client, or a callable consulted before every request.
+
+        Keep-alive sockets belong to the event loop that opened them, so an owner
+        that outlives a loop passes a callable and gets a client per loop. The
+        transport object itself is memoized by callers and cannot be swapped out.
+        """
+        self._resolve_client: Callable[[], httpx.AsyncClient] = (
+            (lambda: client) if isinstance(client, httpx.AsyncClient) else client
+        )
+        self._client = self._resolve_client()
+
+    def _acquire_client(self) -> httpx.AsyncClient:
+        client = self._resolve_client()
+        # `_client` exists for the inherited `_build_request`, which reads client
+        # defaults; keep it on the client this request is actually sent on.
         self._client = client
+        return client
 
     async def send(
         self,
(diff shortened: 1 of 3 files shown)


src/vercel-internal-core/vercel/_internal/core/http/transport.py
```diff
@@ -713,8 +713,28 @@ def __exit__(
...
 
-    def __init__(self, client: httpx.AsyncClient) -> None:
+    def __init__(
+        self,
+        client: httpx.AsyncClient,
+        *,
+        client_provider: Callable[[], httpx.AsyncClient] | None = None,
+    ) -> None:
+        self._client = client
+        self._client_provider = client_provider
```
scotttrinh: Design question:

I personally dislike having two arguments that interact with each other in a non-obvious way. "Does the provider override the client? The other way around? I'm not sure, I have to look at the source or docs.". Could we maybe do:


```suggestion
        client: httpx.AsyncClient | Callable[[], httpx.AsyncClient],
    ) -> None:
        self._client = client
```

And then in the resolve function, check if it's callable?

---

Or just switch to using `client_provider` only since this is an internal API so we don't need to worry about backwards compatability?

kajogo777: agreed

src/vercel-internal-core/vercel/_internal/core/http/transport.py
```diff
@@ -713,8 +713,24 @@ def __exit__(
...
+        """
+        self._resolve_client: Callable[[], httpx.AsyncClient] = (
+            (lambda: client) if isinstance(client, httpx.AsyncClient) else client
+        )
+        self._client = self._resolve_client()
+
+    def _acquire_client(self) -> httpx.AsyncClient:
+        client = self._resolve_client()
+        # `_client` exists for the inherited `_build_request`, which reads client
+        # defaults; keep it on the client this request is actually sent on.
```
scotttrinh: Feel free to change `_build_request` so this is cleaner if you want. It can just be a function that also takes the client instance so you don't have to store the `_client` just for this.

kajogo777: refactored it

scotttrinh (approved):
🏆

PR #208, opened 2026-07-30
Update lock file for new security registry

scotttrinh:
Updates the uv lockfile to resolve Python packages through the new managed Vercel security registry, preserving the existing package versions, hashes, and dependency metadata.


fantix (approved):
LGTM but looks like we need auth in CI. 

Also tagging astral-sh/uv#6349

fantix:
Oh another argument to reject is PR is, I think, that external contributors would struggle with `uv` commands without access to the internal Proxy. Refs astral-sh/uv#6830 for potential `uv-dev-ci.lock`.

scotttrinh:
@fantix We'll work on a workaround for the lockfile churn for our internal development, I'll just close this PR for now until we have a solid idea about how we want to deal with this and/or `uv` allows proxies as a first-class concept in their lockfile format.

PR #209, opened 2026-07-30
Add Python port of the devalue wire format

fantix:
## What

Python port of the JS [`devalue`](https://github.com/sveltejs/devalue) wire format, pinned to upstream **5.9.0**, so the workflow SDK can exchange payloads with the TypeScript SDK. Compatible both directions, including cycles, shared references, and custom reducers/revivers.

Four types carry distinctions Python lacks: `Undefined`, `Hole` (an array hole ≠ a slot holding `undefined`), `JsBigInt(int)` (a `bigint` at any size, `int` subclass so it still computes normally), and `JsRegExp(source, flags)` — kept verbatim rather than compiled, because `re` raises on patterns JS accepts (`(?<n>…)`, `\p{…}`) and silently redefines others (`\d` is ASCII-only in JS).

Untrusted input: an out-of-range index hydrates to `Undefined` rather than wrapping to a real element, and an oversized sparse array is rejected rather than allocated (JS gets this free from dictionary-mode arrays; a Python list is dense).

## Tests

371, in two layers. Unit tests (239) pin exact wire bytes offline. Interop tests (132) run the real JS package via Node — our own values full-circle (Python encode → JS `deepStrictEqual` → JS encode → Python decode), plus **upstream's 106 fixtures and 28 error cases**. Comparison is semantic, not byte-exact; byte equality would need ordered sets and JS number formatting, neither related to correctness.

Worth it: this found a crash where `[["Object",-3]]` (boxed `NaN`) raised `IndexError`.

## Known lossiness

46 upstream fixtures don't round-trip — boxed primitives, typed arrays, `Temporal.*`, `Map`, `URL`, null-prototype objects, invalid `Date`. Recorded as `_KNOWN_LOSSY` and asserted **both** ways, so closing a gap or regressing one each fail until the list is updated. Wrappers get added as we need them.

Most lose only the type tag. **`Map` is the exception**: a `dict` keeps insertion order but not JS key identity, so `1` and `true` collapse into one entry with no error. Pinned separately.

Not ported: `uneval`, `stringifyAsync`, pluggable `operations` — nothing in workflow needs them.

## CI

Adds Node and clones devalue at `v5.9.0` into `RUNNER_TEMP` (upstream's fixtures aren't in the npm tarball). Interop tests skip without Node locally, mandatory in CI, per the `tests/test_examples.py` precedent.

### DIFF
--- a/src/vercel/_internal/devalue/parse.py
+++ b/src/vercel/_internal/devalue/parse.py
@@ -1,0 +1,349 @@
+"""Parse devalue JSON wire format back into Python values."""
+
+from __future__ import annotations
+
+import base64
+import json
+import re as re_module
+from collections.abc import Callable
+from datetime import datetime
+from typing import Any
+
+from .constants import (
+    HOLE,
+    MAX_SPARSE_ARRAY_LENGTH,
+    NAN,
+    NEGATIVE_INFINITY,
+    NEGATIVE_ZERO,
+    POSITIVE_INFINITY,
+    SPARSE,
+    UNDEFINED,
+)
+from .utils import Undefined, is_js_integer, is_valid_array_index, is_valid_array_len
+
+
+def parse(serialized: str, revivers: dict[str, Callable] | None = None) -> Any:
+    """Deserialize a string produced by ``devalue.stringify`` (JS or Python).
+
+    *revivers* is an optional dict mapping custom type tags to functions
+    that receive the hydrated inner value and return the final object.
+    """
+    return unflatten(json.loads(serialized), revivers)
+
+
+def unflatten(parsed: Any, revivers: dict[str, Callable] | None = None) -> Any:
+    """Hydrate an already-JSON-parsed value produced by ``devalue.stringify``.
+
+    Accepts either a negative integer (for special top-level values such as
+    ``undefined`` or ``NaN``) or a list (the flattened representation).
+
+    Input is untrusted: an index that does not address a slot hydrates to
+    ``Undefined`` (as it does in JS), and a sparse array declaring more than
+    ``MAX_SPARSE_ARRAY_LENGTH`` slots is rejected rather than allocated.
+    """
+    if isinstance(parsed, (int, float)) and not isinstance(parsed, bool):
+        return _hydrate_toplevel(parsed)
+
+    if not isinstance(parsed, list) or len(parsed) == 0:
+        raise ValueError("Invalid input")
+
+    values: list[Any] = parsed
+    hydrated: dict[int, Any] = {}
+    hydrating: set[int] | None = None
+
+    def hydrate(index: Any, standalone: bool = False) -> Any:
+        nonlocal hydrating
+
+        # Special sentinel constants
+        if index == UNDEFINED:
+            return Undefined
+        if index == NAN:
+            return float("nan")
+        if index == POSITIVE_INFINITY:
+            return float("inf")
+        if index == NEGATIVE_INFINITY:
+            return float("-inf")
+        if index == NEGATIVE_ZERO:
+            return -0.0
+
+        if standalone or not isinstance(index, (int, float)) or isinstance(index, bool):
+            raise ValueError("Invalid input")
+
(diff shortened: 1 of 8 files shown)


src/vercel/_internal/devalue/parse.py
```diff
@@ -0,0 +1,349 @@
...
+            for i in range(1, len(value), 2):
+                k = hydrate(value[i])
+                v = hydrate(value[i + 1])
+                d[k] = v
+
+        elif type_name == "RegExp":
+            pattern = value[1]
+            flags_str = value[2] if len(value) > 2 else ""
+            py_flags = _js_flags_to_py(flags_str)
+            hydrated[index] = re_module.compile(pattern, py_flags)
```
msullivan: I think maybe we should produce a RegExp container object instead of actually compiling a regex. I don't think the languages match exactly.

src/vercel/_internal/devalue/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    """Format an object key for error path display."""
+    if _IDENTIFIER_RE.match(key):
+        return "." + key
+    return "[" + json.dumps(key) + "]"
+
+
+def is_js_integer(n: object) -> TypeGuard[int | float]:
+    """Mirror of JS ``Number.isInteger``.
+
+    JSON numbers reach Python as ``int`` or ``float``, so ``3.0`` must be
```
msullivan: Sigh, do we really need to support float indexes? Do real JSON serializers produce them? (They shouldn't, in JS!)

msullivan (approved):
One note, and I'm not sure how much we care:
This is lossy in a bunch of places:
* Types that map to str or bytes
* BigInt distinction for ints that don't need to be
* JS can distinguish between an array position having a hole and being undefined

It might be worth adding some wrapper classes to distinguish some of the special types, but I don't feel strongly.

The big potential win could be that if we can losslessly represent all the values, then we could test full round-trippability driven by the upstream test cases, maybe?
Probably not worth the hassle now, though.

src/vercel/_internal/devalue/parse.py
fantix: Agreed! Now fixed as a `JsRegExp(source, flags)` container.

src/vercel/_internal/devalue/utils.py
fantix: You're right! It's now dropped as plain `int` now.

fantix:
Pushed a commit:

**Fixed:**

* `bigint` -> `JsBigInt(int)` subclass, so `1n` survives instead of decaying to 1 (it was inferred from magnitude, hence small ones were lost)
* array holes -> own sentinel, so `[, 1]` no longer re-encodes as `[undefined, 1]`
* `RegExp` -> source+flags container
* bug found: `[["Object",-3]]` — boxed `NaN` — raised `IndexError`; that branch did its own bounds-unchecked `values[...]` lookup. Same for `±Infinity` and `-0`.

**Tests now driven also by upstream cases:**

* 106 fixtures: parse -> re-encode -> compare as values in JS
* 28 error cases: must be rejected identically
* semantic, not byte-exact — the latter would also need ordered sets, JS number formatting and the sparse/`HOLE` heuristic, none related to losslessness

**Still lossy — 46 fixtures, tracked so they are not forgotten**

Recorded as `_KNOWN_LOSSY` keyed by wire tag: boxed primitives, typed arrays + `DataView`, `Temporal.*`, `Map`, `URL`/`URLSearchParams`, null-prototype objects, invalid `Date`. Asserted in **both** directions, so closing a gap fails the test until the entry is dropped, and a regression fails it too. Plan is to add wrappers as we hit them.

Most of those lose only the type tag, not the payload. `Map` is the exception and I'd flag it: a dict keeps insertion order but not JS key identity, so a numeric key can't be re-encoded, an unhashable one can't be parsed, and `1` and `true` collapse into one entry with no error. The upstream fixtures only use string and object keys, so I added tests pinning that separately. Fixing it properly means a pair-sequence container rather than a `dict` - we can add that later.

PR #218, opened 2026-07-31
Add vercel.functions.wait_until()

ricardo-agz:
Expose the runtime's invocation-scoped waitUntil through the SDK: keep the current Function invocation alive for an awaitable that may finish after the response is sent. Outside a Vercel invocation context the call is a safe no-op that closes a lazy coroutine.

Independent of the APScheduler stack (#212–#216); pairs with the runtime support in vercel/vercel (`ricardo/python-wait-until`).

### DIFF
--- a/src/vercel/tests/unit/test_functions_wait_until.py
+++ b/src/vercel/tests/unit/test_functions_wait_until.py
@@ -1,0 +1,42 @@
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+
+from vercel.cache.context import set_context
+from vercel.functions import wait_until
+
+
+@pytest.fixture(autouse=True)
+def clear_wait_until_context() -> Any:
+    set_context(wait_until=None)
+    yield
+    set_context(wait_until=None)
+
+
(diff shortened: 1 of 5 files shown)


src/vercel/tests/unit/test_functions_wait_until.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_wait_until_is_a_noop_without_context() -> None:
+    async def work() -> None:
+        return
+
+    wait_until(work())
+
+
+@pytest.mark.parametrize("value", [None, 1, False, "work", object()])
```
scotttrinh, line 44: what about adding a sync function to this list, since that's probably the most common oopsy.

scotttrinh (approved):
Nice! I guess it's not possible to have sync?

src/vercel/tests/unit/test_functions_wait_until.py
ricardo-agz: yea holding off on adding sync just yet (just for smaller scope), so added a clearer error message for now. but we could probably have sync run in a threadpool later or something

PR #219, opened 2026-07-31
Add project routes API

ricardo-agz:
Add sync and async clients for the project routes API, exposed as
`Vercel.project_routes` and `AsyncVercel.project_routes`.

Everything in and out is a frozen Pydantic model with snake_case fields; an
alias generator handles the camelCase wire format. Changes stage first, and
promoting the staged version publishes them:

```python
from vercel.client import Vercel
from vercel.project_routes import RewriteRoute

vercel = Vercel(access_token="...")

added = vercel.project_routes.add_route(
    project_id="prj_123",
    route=RewriteRoute(name="Rewrite /old to /new", source="/old", destination="/new"),
)

vercel.project_routes.update_route_version(
    project_id="prj_123", version_id=added.version.id, action="promote"
)
```

Results round-trip back into the API. Anything the authoring models
(`RewriteRoute`, `RedirectRoute`, `SetStatusRoute`) don't cover can be passed
as a `vercel.json`-shaped mapping and is validated into a model:

```python
result = vercel.project_routes.get_routes(project_id="prj_123")

vercel.project_routes.stage_routes(
    project_id="prj_123",
    routes=[r for r in result.routes if r.enabled],
    overwrite=True,
)

vercel.project_routes.add_route(
    project_id="prj_123",
    route={
        "name": "Beta users only",
        "srcSyntax": "path-to-regexp",
        "route": {
            "src": "/beta/:path*",
            "dest": "/app/:path*",
            "has": [{"type": "cookie", "key": "beta", "value": "1"}],
        },
    },
)
```

## Notes

- All eight operations are covered, including `edit_route(restore=True)` to
  roll a rule back to production, `get_routes(diff="only")` to preview staged
  changes, and `generate_route`, which returns a `GeneratedRoute` suggestion
  from a prompt and accepts a previous suggestion to refine.
- Server constraints fail fast as `ValueError`: route xor restore,
  before/after placement needs `reference_id`, diff excludes search/filter,
  redirect statuses limited to 301/302/303/307/308.
- Failed requests raise `ProjectRoutesError` with `status_code`, the API error
  `code`, and the parsed body. Response types match the api-project-routes
  schemas.

## Validation

`uv run poe test vercel` (872 passed, 27 skipped), plus lint and typecheck.


scotttrinh (approved):
Looks good! Seems funny to add `iter_coroutine` and our `transport` to this old API, but that works for me anyway.

PR #221, opened 2026-07-31
Release Packages

msullivan:
vercel-internal-core
--------------------

0.1.1 - 2026-07-31
------------------

Bug Fixes
---------

- Bind asynchronous session clients to their running event loop to prevent failures when a shared session is used across multiple event loops or threads. (#220)

vercel-queue
------------

0.7.2 - 2026-07-31
------------------

Bug Fixes
---------

- Don't require `VERCEL_DEPLOYMENT_ID` to be set when running against a `vercel dev` queue broker. (#211)
- This will allow us to revert `vc dev` setting `VERCEL_DEPLOYMENT_ID`. (#211)

vercel-sandbox
--------------

0.3.0 - 2026-07-31
------------------

Features
--------

- Add `get_or_create_sandbox` to retrieve a named sandbox or create it when it does not exist, for both async and synchronous APIs. (#220)

vercel
------

0.8.0 - 2026-07-31
------------------

Breaking Changes
----------------

- `vercel.sandbox` is now the promoted Sandbox API that previously lived at `vercel.unstable.sandbox`. The former `vercel.sandbox` surface, including `AsyncSandbox`, `Command`, `AsyncCommand`, and `TokenProvider`, is gone, and `vercel.unstable` has been removed. Use `vercel.sandbox` for the async API and `vercel.sandbox.sync` for the synchronous API. (#195)
- The Sandbox implementation now ships in the separate `vercel-sandbox` distribution, which `vercel` depends on, so `vercel.sandbox` imports keep working without installing anything extra. (#195)
- Workflow and step identifiers now use `//` between the module and qualified name to match the TypeScript SDK format. (#220)

Features
--------

- Add `vercel.functions.wait_until()` for post-response asynchronous work that remains attached to the current Python Function invocation. (#218)
- Port from vercel-workers to vercel-queue (#184)

Internal
--------

- Move Vercel SDK tests under the package-local test suite. (#194)

vercel-celery
-------------

0.7.2 - 2026-07-31
------------------

- Update dependencies.

vercel-dramatiq
---------------

0.7.1 - 2026-07-31
------------------

- Update dependencies.


fantix (approved):

src/vercel-sandbox/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.3.0 - 2026-07-31
```
scotttrinh, line 3: Do feature-only bumps end up as minor bumps even under 0.x semver rules?

scotttrinh (approved):

src/vercel-sandbox/CHANGELOG.md
msullivan: They probably don't need to, but I think that's what the release system does now

PR #224, opened 2026-08-01
workflow: make payloads interchangeable with the TypeScript SDK

fantix:
Run inputs, step inputs and results, workflow outputs and hook payloads went out as `b"json"` plus `json.dumps`, wrapped in a one-element list — neither half is anything `@workflow/core` writes. They are now devalue payloads tagged `devl`,
stored as a single value, and a call is recorded as the single object TypeScript records for a one-argument call, so a Python run reads the same as one the TypeScript SDK started.

devalue only carries what JavaScript has a form for, so `Decimal`, `UUID`, `date`, `time`, `timedelta` and `Path` ride `@workflow/core`'s `Instance` tag, and `@serializable` puts your own classes there — a JavaScript reader that has never heard of the class still renders the value rather than losing the payload around it.

## Breaking

- **Workflows and steps take keyword arguments only**, and their parameters have to be declared keyword-only — `async def charge(*, amount: int)`. Every definition needs updating; a positional parameter is rejected when the function is registered, so a codebase reports all of them at import rather than one per run. Calls become `start(wf, amount=21)` / `charge(amount=21)`.
- **Runs recorded by an earlier version cannot be replayed.** Their payloads carry the old `json` prefix and are rejected with a `SerializationError`. In-flight runs in local dev projects should be drained before deploying this.
- **`BaseHook.resume()` forwards its keyword arguments to `model_dump()`** rather than `model_dump_json()`, in `"python"` mode. JSON-only options such as `indent` are no longer accepted; a `datetime` field now reaches the workflow as a `datetime` instead of a string; and a field whose class nothing has registered raises rather than being flattened to a string.

Verified end to end against the published `workflow` CLI, on the local world and on production: `input: [ { sku: 'ABC' } ]`, `output: { total: Decimal@decimal '1.50', … }`. Best read commit by commit — the first is a packaging fix (`_internal/devalue` was never in the wheel, which the second would have made fatal at import).

### DIFF
--- a/changes/vercel/20260731134500.breaking.md
+++ b/changes/vercel/20260731134500.breaking.md
@@ -1,0 +1,11 @@
+Workflow payloads — run and step inputs, step results, workflow outputs and
+hook payloads — now use the devalue wire format of the TypeScript
+`@workflow/core` package: a `devl` prefix, and a single value per field where
+this SDK wrote a one-element list. Workflows and steps are also called with
+keyword arguments only, and their parameters have to be declared keyword-only
+(a positional one is rejected when the function is registered). Together those
+make a Python call record byte-identically to the TypeScript one, so payloads
+written by either SDK are readable by the other. Runs recorded by an earlier
+version cannot be replayed. `BaseHook.resume()` now passes its keyword
+arguments to `model_dump()`, in `"python"` mode, rather than to
+`model_dump_json()`.
--- a/changes/vercel/20260731134500.breaking.md
+++ b/changes/vercel/20260731134500.breaking.md
@@ -1,11 +1,3 @@
-Workflow payloads — run and step inputs, step results, workflow outputs and
-hook payloads — now use the devalue wire format of the TypeScript
-`@workflow/core` package: a `devl` prefix, and a single value per field where
-this SDK wrote a one-element list. Workflows and steps are also called with
-keyword arguments only, and their parameters have to be declared keyword-only
-(a positional one is rejected when the function is registered). Together those
-make a Python call record byte-identically to the TypeScript one, so payloads
-written by either SDK are readable by the other. Runs recorded by an earlier
-version cannot be replayed. `BaseHook.resume()` now passes its keyword
-arguments to `model_dump()`, in `"python"` mode, rather than to
-`model_dump_json()`.
+Workflow payloads now use the devalue wire format of the TypeScript
+`@workflow/core` package, and workflows and steps are called with keyword
+arguments only.
--- a/src/vercel/_internal/workflow/serde.py
+++ b/src/vercel/_internal/workflow/serde.py
@@ -28,14 +28,12 @@ TypeScript compiler plugin generates.
 
 from __future__ import annotations
 
+import contextlib
+import contextvars
 import dataclasses
-import datetime
-import decimal
 import enum
 import operator
-import pathlib
-import uuid
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from typing import Any, TypeVar
 
 CLASS_ID_PREFIX = "class//"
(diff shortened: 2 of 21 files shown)


src/vercel/_internal/workflow/serde.py
```diff
@@ -0,0 +1,285 @@
...
+    *,
+    class_id: str | None = None,
+    serialize: Callable[[T], Any] | None = None,
+    deserialize: Callable[[Any], T] | None = None,
+) -> None:
+    """Teach the wire format about *cls*.
+
+    *serialize* turns an instance into something devalue can already carry;
+    *deserialize* turns that back into an instance. Both default to the
+    ``__workflow_serialize__`` / ``__workflow_deserialize__`` methods on the
```
msullivan: It might be better to just do `_workflow_serialize`, etc.
In theory, dunders are supposed to be reserved for python itself.
(Though certainly we violated that rule plenty in edgedb-python.)

src/vercel/_internal/workflow/serde.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+_by_class_id: dict[str, _Registration] = {}
+_by_class: dict[type, _Registration] = {}
+# Types devalue already carries that a registration on a base class would
+# otherwise capture. Walking the MRO stops here, declining the value.
+_NATIVE: set[type] = {datetime.datetime}
+# `type(value)` -> the registration that applies to it, or None. Cleared
+# whenever a registration lands, since a new one can change the answer.
+_resolved: dict[type, _Registration | None] = {}
```
msullivan: Depending on how we are expecting this to be used, we should consider whether we want weak maps here. If code loaded in the sandbox might register stuff...
(We could also have a sandbox cleanup that prunes stuff from this?)

src/vercel/_internal/workflow/serde.py
```diff
@@ -0,0 +1,285 @@
...
+        return registration.deserialize(value.get("data"))
+    except Exception as error:
+        # Payload data is untrusted, and a registered deserializer may raise
+        # anything at all -- `Decimal("garbage")` raises `InvalidOperation`,
+        # which is not even an error the codec would recognize.
+        raise ValueError(f"{class_id} could not be read back: {error}") from error
+
+
+def registration_hint(value: Any) -> str:
+    """A sentence naming the way out, for a value nothing can serialize."""
```
msullivan: lol why are the robots like this

changes/vercel/20260731134500.breaking.md
```diff
@@ -0,0 +1,11 @@
...
+hook payloads — now use the devalue wire format of the TypeScript
+`@workflow/core` package: a `devl` prefix, and a single value per field where
+this SDK wrote a one-element list. Workflows and steps are also called with
+keyword arguments only, and their parameters have to be declared keyword-only
+(a positional one is rejected when the function is registered). Together those
+make a Python call record byte-identically to the TypeScript one, so payloads
+written by either SDK are readable by the other. Runs recorded by an earlier
+version cannot be replayed. `BaseHook.resume()` now passes its keyword
+arguments to `model_dump()`, in `"python"` mode, rather than to
+`model_dump_json()`.
```
msullivan: Clean this up a bit to make the different parts more clear.

msullivan (approved):
Please clean up the commit message to make the rationale for various decisions more clear.

Also let's try to make sure that we could start doing positional arguments also again without needing another compatibility break.
(I'm somewhat ambivalent about this, but JS workflow can do positional args, right?)

src/vercel/_internal/workflow/serde.py
fantix: Make sense, now fixed.

src/vercel/_internal/workflow/serde.py
fantix: This actually leads to an escaping issue - the `run_workflow()` input hydration should happen inside the `workflow_sandbox()`. Now fixed the `serde` with `ContextVar`-based layered registration.

src/vercel/_internal/workflow/serde.py
fantix: English! Reworded anyways

changes/vercel/20260731134500.breaking.md
fantix: Sure!

src/vercel/_internal/workflow/serde.py
fantix: [66bac54b](https://github.com/vercel/vercel-py/pull/224/changes/66bac54bc2ec603e8a1f357e2172d9aabfb2c544)

fantix:
> Please clean up the commit message to make the rationale for various decisions more clear.

The already-reviewed 4 commits are unchanged, but with refreshed commit messages. I added new commits for easier reviewing.

> Also let's try to make sure that we could start doing positional arguments also again without needing another compatibility break.
>
> (I'm somewhat ambivalent about this, but JS workflow can do positional args, right?)

Yes the JS workflow can do positional args, and only positionals actually. This PR has compatible wire format if we decide to allow positional args.

msullivan (approved):

PR #229, opened 2026-08-04
workflow: Start using a custom event loop for workflows

msullivan:
The advantages here are:
1. We'll be able to ensure that no I/O can be done on the loop,
   even from libraries
2. We can get rid of other hacks

Currently this is done in a pretty minimal way by overloading BaseEventLoop.
Eventually we'll probably end up building our complete own version.

---


fantix (approved):
Love it!

PR #234, opened 2026-08-04
Add package-owned example test tasks

scotttrinh:
Move example execution into package-owned Poe tasks and add an opt-in workspace aggregate with scoped passthrough.


msullivan (approved):

fantix (approved):
LGTM!

PR #236, opened 2026-08-04
Add sandbox auto-resume and session scopes

scotttrinh:
Adds automatic Sandbox session recovery for sync and async operations. When a covered process, snapshot, network-policy, execution-limit, or filesystem operation encounters a stopped session, the existing handle resumes the sandbox and replays the operation once against the replacement session.

Recovery coordinates concurrent callers so they share one resume attempt, polls bounded stopping or snapshotting transitions, and preserves cancellation and interruption behavior. Lazy handles follow the active session, while runtime-session handles, process handles, and already-open streams remain pinned to their original session.

Adds `box.session()` scopes for acquiring an active session and stopping that exact session on exit. Cleanup preserves replacement sessions and sparse sandbox metadata during recovery races.

Filesystem recovery covers reads, writes, batch writes, and unsized writers while retaining at-least-once write semantics. Failures after streaming begins are never replayed. Recovery also preserves route projection, session identity, HTTP 410 handling, and the one-recovery/one-replay limit.

### DIFF
--- a/src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
+++ b/src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
@@ -21,6 +21,7 @@ from vercel.sandbox._internal.async_filesystem_handle import (
     SandboxTextWriter,
 )
 from vercel.sandbox._internal.errors import (
+    SandboxApiError,
     SandboxCleanupError,
     SandboxResponseError,
     SandboxTerminalStateError,
@@ -70,6 +71,7 @@ from vercel.sandbox._internal.recovery import (
     TRANSITION_TIMEOUT,
     SandboxLifecycle,
     SandboxRecoveryTarget,
+    classify_sandbox_lifecycle_error,
     execute_with_sandbox_recovery,
 )
 from vercel.sandbox._internal.runtime_common import (
@@ -846,7 +848,7 @@ class SandboxRuntimeSession(RuntimeSessionHandleBase):
         payload = await self._service.get_runtime_session(
             session_id=self.id, include_system_routes=include_system_routes
         )
-        self._apply_payload(payload)
+        self._apply_payload_with_parent(payload)
         return self
 
     async def extend_execution_time_limit(self, duration: DurationInput) -> Self:
@@ -885,8 +887,8 @@ class SandboxRuntimeSession(RuntimeSessionHandleBase):
 
     async def stop(self) -> Self:
         """Stop this runtime session and refresh the handle."""
-        payload = await self._service.stop_runtime_session(session_id=self.id)
-        self._apply_payload(payload)
+        result = await self._service.stop_runtime_session(session_id=self.id)
+        self._apply_stop_result(result)
         return self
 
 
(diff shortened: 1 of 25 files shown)


src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
```diff
@@ -862,21 +870,56 @@ def __init__(
...
             project_id=self.project_id,
             include_system_routes=self._include_system_routes,
         )
         self._apply_payload(payload)
+
+    def _clear_recovery_task(self, task: asyncio.Task[None]) -> None:
+        if self._recovery_task is task:
+            self._recovery_task = None
+        if not task.cancelled():
+            task.exception()
```
fantix: Indent these 2 lines so that it only raises when `self._recovery_task is task`?

src/vercel-sandbox/vercel/sandbox/_internal/sync_runtime.py
fantix: This should mirror the async `_apply_current_session_payload_if_current()`?

src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
```diff
@@ -899,6 +948,11 @@ async def _await_shared_resume(self) -> None:
             task.add_done_callback(self._clear_recovery_task)
         await asyncio.shield(task)
 
+    async def _ensure_active(self) -> str:
+        """Ensure an active current session before binding a lazy stream."""
+        await self._await_shared_resume()
```
fantix: I was expecting something like this here:

```python
if not active:
    await self._await_shared_resume()
```

Or even calling `_wait_for_transition()`?

fantix (approved):

src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
scotttrinh: I'm glad you brought this up. While looking through this more closely, I realized I had messed up here in general: we are using AnyIO explicitly in this project, and their structured way of doing this kind of multi-waiter architecture is quite a bit different than the thing we put together here. I was trying to keep abandoned resumes alive since they carry useful state data in the response even if we don't need the new session anymore, but that's not worth the large amount of bookkeeping that would be required to get AnyIO to be happy with this.

Will fix!

src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
scotttrinh: True. I've updated `_ensure_active` to just delegate to the existing `_acquire_session` flow which makes this align with the existing behavior elsewhere.

src/vercel-sandbox/vercel/sandbox/_internal/sync_runtime.py
scotttrinh: Yep, good catch here, too. It now retains it's snapshot ID and only applies the response if that session is still current.

src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
scotttrinh: https://github.com/vercel/vercel-py/pull/236/commits/c313b124a5d18ede8c60b805cf0784f5f1632a7e

Take a close look at this to make sure this still makes sense. Instead of the abandoned task updating the sandbox, it now truly abandons if every one of the waiters cancels. It's a behavioral change, but I wasn't married to the original design anyway, just thought it might be useful to get updated state, but this is easier to reason about, for sure.

fantix: Love it! And it's more consistent with the blocking end, LGTM!

fantix (approved):

PR #243, opened 2026-08-05
workflow: record positional arguments again

fantix:
A workflow/step call is bound to the callee's signature and recorded as the argument array JS `@workflow/core` writes, with keyword arguments in a trailing object. Named parameters (including those positional arguments that can also be set as kwarg) go by name, so a reorder is safe and a rename is loud; positional-only (the args before `/`) and `*args` go by position.


src/vercel/_internal/workflow/serialization.py
```diff
@@ -86,53 +87,84 @@ def hydrate(data: Any, *, what: str) -> Any:
...
+    also carries keyword arguments, and the array has one slot per value with
+    nowhere to name anything — so the keyword arguments ride in a single object
+    appended after the positional ones. Takes the two halves
+    `core._bind_arguments` produced, not a raw call:
+
+    ================  ==========================  ==========================
+    positional        keyword                     array
+    ================  ==========================  ==========================
+    ``()``            ``{}``                      ``[]``
+    ``(21,)``         ``{}``                      ``[21]``
```
msullivan, line 114: I feel like my inclination would be to, for simplicity and consistency, always emit `[21, {}]` here. (That is, unconditionally add the kwargs empty dict.)

But it *is* a (3 byte) optimization to not do that, and has advantages of being a little cleaner when it *is* all scalar positional args, so I guess we probably should just keep it.

src/vercel/_internal/workflow/core.py
```diff
@@ -25,30 +25,74 @@
...
+def _bind_arguments(
+    signature: inspect.Signature,
+    qualname: str,
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+) -> tuple[tuple[Any, ...], dict[str, Any]]:
+    """Split a call by what the callee's parameters can be *named*.
+
+    A value whose parameter has a usable name is recorded by name; only a
+    parameter that cannot be named -- positional-only (before a `/`) or
```
msullivan, line 37: Will biasing towards keyword args fit right with your plans for JS compat testing?
(I guess it's easy to change if needed)

msullivan (approved):
Looks good!

src/vercel/_internal/workflow/core.py
fantix: For e2e test, I believe the input wire format written by Python should be Python implementation detail, as far as Python could read and apply the JS-written positional-only input args.

For future interoperability, we'll need a wrapper codegen kinda mechanism anyways, so the codegen should be able to digest that bias.

src/vercel/_internal/workflow/serialization.py
fantix: Yeah idk which is better - but ig the wire format is compatible, like if we decide to add the trailing empty dict in the future, the code can still read the old `[21]` fwiw and apply properly.

PR #245, opened 2026-08-05
Add Runtime Cache backend for APScheduler integration

nsidnev:
Adds another backend for APScheduler integration based on Vercel Runtime Cache.

---

Redis backend stays the fully-guaranteed backend, but it is no longer the entry fee. A project with no `RedisJobStore` now runs on the Vercel Runtime Cache:

```python
scheduler = BlockingScheduler()   # no jobstores, no REDIS_URL

@scheduler.scheduled_job("cron", hour=4, id="cleanup", replace_existing=True)
def cleanup() -> None: ...
```

```toml
[[tool.vercel.subscribers]]
entrypoint = "scheduler:scheduler"
```

Backend selection is automatic (a configured `RedisJobStore` selects redis; anything else selects cache) and overridable with `VERCEL_APSCHEDULER_BACKEND=redis|cache`. The Redis path is byte-for-byte the existing behavior.

The integration core (_adapter, _subscriber, payloads) no longer contains backend-specific code. Shared contract types moved to _types.py; the Redis driver and coordinator moved (git mv, history preserved) under `_backends/redis/`; the new backend lives under `_backends/cache/`. Backends satisfy three protocols — `Driver`, `JobCoordinator`, `Backend`

---

### How it works

The Runtime Cache has no compare-and-swap and its entries are evictable, so none of the Redis invariants can live in storage. Each guarantee is relocated to something that can actually carry it:

```
in redis mode, the store is the authority and messages point into it
in cache mode, the messages are the authority and the store is a hint
```

- Single successor moves to the queue. Racing finishers compute identical canonical logical times and identical idempotency keys (`aps:wake:{scope}:{sid}:{gen}:{seq}`); the queue accepts the publication once. The chain cannot fork even when claims race. Claims themselves become best-effort filters, so the contract is `at-least-once` with a wider duplicate window than redis mode.
- Chain progress travels in the messages. A claim adopts any generation or wake token strictly ahead of the local document, so an evicted document never strands the chain. `paused` fences only its own and older generations; a `resume's new generation revives a paused document.
- Declared jobs are reconstructable, not durable. Code is the backup: takeover reconciliation doubles as the eviction repairer, rewriting the declared set (provenance-tagged, fingerprint-compared) whenever the reconciled marker is missing. A fully evicted namespace heals on the next wake.
- Runtime mutations and lifecycle flags are best-effort by design: `read-merge-write` with bounded jittered retries, last writer wins, lost on eviction. `pause()` additionally rides the queue as a control message on the start topic, because a cache write is only as visible as the cache — the queue is the one bus both backends share.

Comparasion table:

| Property            | Redis                 | Runtime Cache                                  |
|---------------------|-----------------------|------------------------------------------------|
| One chain, no forks | atomic Lua claims     | queue idempotency keys                         |
| Execution           | at-least-once         | at-least-once, wider duplicate window          |
| Declared jobs       | durable               | rebuilt from code after eviction               |
| Runtime add_job()   | durable, revision-CAS | best-effort                                    |
| pause()             | durable, fails closed | best-effort flag + queue-borne control message |

---

### Adapter changes

Two deliberate behavior changes while wiring the seam. Reconciliation gating is now durable-marker-first (`reconciled_deployment` consulted before the in-process flag, which becomes a cache of it) — required for eviction self-heal, and it also fixes re-reconciliation when a rollback lands on a warm instance. And `_scope_outlives_deployments` is always true for the cache backend: its documents are evictable in every scope, so reconciliation-from-code applies even to deployment-scoped previews.

### DIFF
--- a/integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
+++ b/integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
@@ -1,0 +1,102 @@
+"""Runtime Cache backend: redis-less coordination with documented tradeoffs.
+
+The Vercel Runtime Cache offers ``get``/``set``/``delete`` with TTLs and tags
+— no compare-and-swap, no transactions, and entries may be evicted. This
+backend therefore divides responsibility differently from Redis:
+
+- **Chain integrity** does not live here at all. The single-successor
+  guarantee comes from the queue's idempotency keys: racing finishers compute
+  identical successor payloads (logical times are canonical), and the queue
+  accepts the publication once. Claims are best-effort filters that shrink,
+  but cannot eliminate, duplicate wake *executions*; the contract is
+  at-least-once with a wider duplicate window than Redis mode.
+- **Declared jobs are reconstructable, not durable.** Code is the backup: a
+  missing driver or job document is rebuilt from declarations by the existing
+  reconcile/materialize machinery, so eviction can only cost state that code
+  cannot restate (runtime-added jobs, lifecycle flags).
+- **Runtime mutations and lifecycle flags are best-effort** by declared
+  policy: read-merge-write with bounded retries; last writer wins. ``pause()``
+  additionally rides the queue as a control message.
+
+Outside deployed environments (``vercel dev``) the cache client falls back to
+per-process memory, which makes this backend the in-memory dev mode with the
+queue-serving sidecar as the effective scheduler process.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from dataclasses import dataclass
+from os import environ
+
+from ..._options import SUBSCRIBER_ID_ENV, _SchedulerIdentity
+from ..._types import APSchedulerConfigurationError
+from .._protocols import Driver, JobCoordinator
+from ._driver import CacheDriver
+from ._jobstore import CacheJobCoordinator, CacheJobStore
+
+__all__ = ["CacheBackend", "CacheDriver", "CacheJobCoordinator", "CacheJobStore"]
+
+
+@dataclass
+class _Bound:
+    driver: Driver
+    coordinator: JobCoordinator
+
+
+class CacheBackend:
+    name = "cache"
+
+    def validate_configuration(self, scheduler: Any) -> dict[str, Any]:
+        stores = dict(scheduler._jobstores)
+        extra = set(stores) - {"default"}
+        if extra:
+            aliases = ", ".join(sorted(extra))
+            raise APSchedulerConfigurationError(
+                "vercel-apscheduler v1 supports exactly one job store named "
+                f'"default"; configured: {aliases}'
+            )
+        default = stores.get("default")
+        if default is not None and not isinstance(default, CacheJobStore):
+            raise APSchedulerConfigurationError(
+                f'job store "{type(default).__name__}" is not suitable for the '
+                "cache backend, which injects its own store; remove the "
+                "explicit default store, or select the matching backend via "
+                "VERCEL_APSCHEDULER_BACKEND"
+            )
+        return stores
+
+    def supports_store(self, store: Any) -> bool:
+        return isinstance(store, CacheJobStore)
+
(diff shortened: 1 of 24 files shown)


integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+    def supports_store(self, store: Any) -> bool:
+        return isinstance(store, CacheJobStore)
+
+    def identity_ready(self, scheduler: Any) -> bool:
+        return True
+
+    def derive_identity(self, scheduler: Any) -> _SchedulerIdentity:
+        # No store key to derive from; the builder-assigned subscriber id is
+        # the durable identity, "default" outside Vercel.
+        subscriber_id = environ.get(SUBSCRIBER_ID_ENV) or "default"
```
ricardo-agz, line 79: this identity would differ between the web function that starts the scheduler and the subscriber scheduler function itself since the env var is only available in the subscriber

maybe we could derive this identity from `VERCEL_APSCHEDULER_SUBSCRIBERS`?

integrations/vercel-apscheduler/vercel/integrations/apscheduler/_subscriber.py
```diff
@@ -167,6 +173,26 @@ def _process_wakeup(
...
+    message: vqs.Message[dict[str, Any]],
+) -> None:
+    """Apply a queue-borne lifecycle flag to the local driver document.
+
+    Only the cache backend publishes these; pausing an already-paused chain
+    is a no-op, so redelivery needs no dedup.
+    """
+    if not _targets_adapter(adapter, payload.scheduler_id, message, "lifecycle"):
+        return
+    if payload.action == "pause":
```
ricardo-agz, line 190: could there be a race here where `pause()` and then `resume()` settles as paused? 
maybe we need to gate this so it doesn't pause a schedule that was activated after the pause was issued?

integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/_driver.py
```diff
@@ -0,0 +1,519 @@
...
+        doc = self._read()
+        state = doc.get("state")
+        return DriverSnapshot(
+            state=_lifecycle_state(state) or "inactive",
+            generation=int(doc.get("generation") or 0),
+            start_status=doc.get("start_status"),
+            current_wake=self._token(doc),
+        )
+
+    def repair_overdue_wake(self, now: datetime) -> WakeToken | None:
```
ricardo-agz, line 491: is there a case where we could have a dead start that we need to repair? as opposed to only needing to repair dead wakes?

integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+— no compare-and-swap, no transactions, and entries may be evicted. This
+backend therefore divides responsibility differently from Redis:
+
+- **Chain integrity** does not live here at all. The single-successor
+  guarantee comes from the queue's idempotency keys: racing finishers compute
+  identical successor payloads (logical times are canonical), and the queue
+  accepts the publication once. Claims are best-effort filters that shrink,
+  but cannot eliminate, duplicate wake *executions*; the contract is
+  at-least-once with a wider duplicate window than Redis mode.
+- **Declared jobs are reconstructable, not durable.** Code is the backup: a
```
ricardo-agz, line 13: do we need to handle the case of interval triggered schedules with no explicit start_date? to make them deterministic and not dependent on import time?

ricardo-agz (approved):

integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
nsidnev: yes, true. this is fixed in the next PR of the stack: https://github.com/vercel/vercel-py/pull/253

PR #241, opened 2026-08-05
Release Packages

ricardo-agz:
vercel-oidc
-----------

0.8.0 - 2026-08-05
------------------

Features
--------

- Add OIDC token signature verification. `vercel.oidc.verify_vercel_oidc_token` and its async twin `vercel.oidc.aio.verify_vercel_oidc_token` verify a Vercel OIDC token against the JWKS at `oidc.vercel.com`, pinning the issuer, requiring RS256, and checking the project, environment, owner, and audience claims. `vercel.oidc.extract_bearer_token` reads the credential out of request headers. Verification fails closed: when the expected project or environment cannot be resolved from the arguments or the environment, every token is rejected. (#205)
- Claims are compared for equality only. There is no project wildcard: `"*"` is an ordinary string, so a token cannot widen its own scope to every project in a team. (#205)
- The issuer is pinned to Vercel's OIDC service and is not configurable. Both the root issuer `https://oidc.vercel.com` and the team-scoped `https://oidc.vercel.com/<team>` are accepted, since Vercel mints both and one global key signs them; the JWKS URL is a constant, so a token can never influence where signing keys come from. (#205)
- `vercel.oidc.resolve_vercel_oidc_token_identity`, and its async twin, return an opaque, stable identity for a token. A token is a signature over an identity plus an expiry, so one identity is issued many tokens over time; this is what to key identity-scoped client state on. The signature, issuer and expiry are verified before anything is read, but no claim is checked and none is returned, so it is not an authorization check. (#205)
- This requires the new `verify` extra, which pulls in `pyjwt[crypto]`: (#205)
- pip install "vercel-oidc[verify]" (#205)
- The extra keeps `cryptography` off installs that do not verify tokens. (#205)

vercel-connect
--------------

0.1.0 - 2026-08-05
------------------

Features
--------

- Add `vercel.connect`, a Python SDK for Vercel Connect: short-lived third-party credentials brokered through the deployment's Vercel OIDC identity, with token caching, authorization flows, connector metadata, and inbound trigger verification. (#205)

vercel-internal-telemetry
-------------------------

0.7.2 - 2026-08-05
------------------

- Update dependencies.

vercel-queue
------------

0.7.3 - 2026-08-05
------------------

- Update dependencies.

vercel-sandbox
--------------

0.3.1 - 2026-08-05
------------------

Internal
--------

- Run Sandbox examples through the package-owned workspace Poe task. (#234)

vercel-cache
------------

0.7.2 - 2026-08-05
------------------

[...]

PR #247, opened 2026-08-06
Release Packages

ricardo-agz:
Regenerated release changeset from current main. Replaces #241, whose snapshot predates the APScheduler stack. First release of `vercel-apscheduler` (seeded to 0.0.0 so the tooling computes 0.1.0).

Publishing note: the `vercel-apscheduler-bundle` trusted publisher cannot be registered until `vercel-apscheduler` exists on PyPI (pending-publisher configs must be unique), so the first publish run is expected to fail at the bundle upload. After registering the bundle publisher, resume with `publish.yml` dispatch and `force: true`.

# vercel

### Breaking Changes

- The local workflow world now stores its `.workflow-data` files as JSON in the same format the TypeScript `@workflow/world-local` package uses, instead of CBOR. Runs, steps, hooks and events written by either SDK are now readable by the other. Existing `.workflow-data` directories are not readable in the new format and should be deleted. (#226)
- Workflow payloads now use the devalue wire format of the TypeScript `@workflow/core` package, and workflows and steps are called with keyword arguments only. (#224)

### Features

- Workflow payloads can now carry native `Decimal`, `UUID`, `date`, `time`, `timedelta` and `Path`, and `@serializable` (or `register_serializable()`) is offered for custom classes. (#224)
- Add sync and async clients with typed models for managing project-level routing rules and versions. (#219)

### Bug Fixes

- Allow workflows on Python 3.12 and earlier to import `uuid` by safely exposing `platform.system()` while continuing to block host-specific platform inspection. (#242)

### Internal

- Use a consistent isolated event-loop lifecycle for workflow execution on Python 3.10. (#242)
- Run workflows on a dedicated event loop that advances execution when the loop becomes idle. (#242)
- Avoid invoking the workflow event loop's idle hook after the loop begins stopping. (#242)

# vercel-apscheduler

### Features

- Add the durable Redis driver for running APScheduler schedules through delayed Vercel Queue messages. (#242)
- Add Redis-backed APScheduler subscribers for Vercel Queues. The integration patches `scheduler.start()`, `scheduler.pause()`, and `scheduler.resume()` with durable, deployment-scoped lifecycle transitions and atomic single-chain fencing. Paused occurrences are skipped on resume, and interrupted successor publication is repaired on retry. Production schedules activate on the first request, and opted-in previews stop after a durable idle timeout. Jobs that do not choose a `misfire_grace_time` run their occurrences whenever the wake arrives: the stock one-second grace assumes in-process wakeup precision that queue delivery cannot meet. A takeover atomically fences the demoted deployment's job-store writes, and reconciliation retries revision races, marking completion only once it converges. Automatic takeover reacts to the first request that arrives through an environment alias, regardless of what arrived before it. (#238)

# vercel-oidc

### Features

- Add OIDC token signature verification. `vercel.oidc.verify_vercel_oidc_token` and its async twin `vercel.oidc.aio.verify_vercel_oidc_token` verify a Vercel OIDC token against the JWKS at `oidc.vercel.com`, pinning the issuer, requiring RS256, and checking the project, environment, owner, and audience claims. `vercel.oidc.extract_bearer_token` reads the credential out of request headers. Verification fails closed: when the expected project or environment cannot be resolved from the arguments or the environment, every token is rejected. (#205)
- Claims are compared for equality only. There is no project wildcard: `"*"` is an ordinary string, so a token cannot widen its own scope to every project in a team. (#205)
- The issuer is pinned to Vercel's OIDC service and is not configurable. Both the root issuer `https://oidc.vercel.com` and the team-scoped `https://oidc.vercel.com/<team>` are accepted, since Vercel mints both and one global key signs them; the JWKS URL is a constant, so a token can never influence where signing keys come from. (#205)
- `vercel.oidc.resolve_vercel_oidc_token_identity`, and its async twin, return an opaque, stable identity for a token. A token is a signature over an identity plus an expiry, so one identity is issued many tokens over time; this is what to key identity-scoped client state on. The signature, issuer and expiry are verified before anything is read, but no claim is checked and none is returned, so it is not an authorization check. (#205)
- This requires the new `verify` extra, which pulls in `pyjwt[crypto]`: (#205)
- pip install "vercel-oidc[verify]" (#205)
- The extra keeps `cryptography` off installs that do not verify tokens. (#205)

### Bug Fixes

- Record JWKS refetch outcomes before allowing another caller to fetch, avoiding duplicate requests under concurrency. (#242)

# vercel-connect

### Features

- Add `vercel.connect`, a Python SDK for Vercel Connect: short-lived third-party credentials brokered through the deployment's Vercel OIDC identity, with token caching, authorization flows, connector metadata, and inbound trigger verification. (#205)

# vercel-sandbox

### Breaking Changes

- Require synchronous credential factories when configuring `vercel.sandbox.sync`; use asynchronous factories only with the async Sandbox API. (#242)

### Internal

- Run Sandbox examples through the package-owned workspace Poe task. (#234)

# vercel-internal-core

### Internal

- Key session service options by logical service so synchronous and asynchronous variants share configuration safely. (#242)

# vercel-internal-telemetry

## 0.7.2 - 2026-08-05

- Update dependencies.

# vercel-queue

## 0.7.3 - 2026-08-05

- Update dependencies.

# vercel-cache

## 0.7.2 - 2026-08-05

- Update dependencies.

# vercel-celery

## 0.7.3 - 2026-08-05

- Update dependencies.

# vercel-dramatiq

## 0.7.2 - 2026-08-05

- Update dependencies.


nsidnev:
Superseded by https://github.com/vercel/vercel-py/pull/268

PR #248, opened 2026-08-06
workflow: honour the workflow-server and proxy URL overrides

fantix:
`VercelWorld` hard-coded both hops to the workflow API, so two documented environment variables did nothing in Python:

- `VERCEL_WORKFLOW_SERVER_URL` — the workflow-server itself.
- `WORKFLOW_VERCEL_BACKEND_URL` — the api.vercel.com proxy in front of it.

They are how a deployment gets pointed at something other than production `vercel-workflow.com`. Ignoring them means a Python app writes its runs to a different server than the rest of the toolchain reads them from.

```console
$ VERCEL_WORKFLOW_SERVER_URL=https://workflow-server-git-branch.vercel.sh uv run python -c "
from vercel._internal.workflow.worlds.vercel import VercelWorld
print(VercelWorld(token='t')._base_url)
print(VercelWorld(token='t', project_id='prj_1', team_id='team_1')._headers['x-vercel-workflow-api-url'])
"
https://workflow-server-git-branch.vercel.sh/api    # was https://vercel-workflow.com/api
https://workflow-server-git-branch.vercel.sh        # was absent
```

Also drops the unused `WORKFLOW_SERVER_URL_OVERRIDE` constant. It mirrored an inline escape hatch in `world-vercel`, which is only safe there because a lint job fails the build unless it is empty; with no such guard here, a hand-edited value could reach `main` unnoticed.

7 new tests; `uv run poe qa vercel` green.

### DIFF
--- a/src/vercel/_internal/workflow/worlds/vercel.py
+++ b/src/vercel/_internal/workflow/worlds/vercel.py
@@ -17,25 +17,6 @@ from vercel.oidc.aio import get_vercel_oidc_token
 
 from .. import world as w
 
-# Hard-coded workflow-server URL override for testing.
-# Set this to test against a different workflow-server version, or leave it
-# empty and set VERCEL_WORKFLOW_SERVER_URL in the environment instead.
-# Empty and unset means production (uses default vercel-workflow.com).
-#
-# Example: 'https://workflow-server-git-branch-name.vercel.sh'
-#
-WORKFLOW_SERVER_URL_OVERRIDE = ""
-
-
-def _workflow_server_url_override() -> str:
-    """Resolve the workflow-server URL override, or "" for the default.
-
-    Mirrors world-vercel's precedence in utils.ts: the hard-coded constant
-    wins, then VERCEL_WORKFLOW_SERVER_URL.
-    """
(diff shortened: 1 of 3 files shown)


src/vercel/_internal/workflow/worlds/vercel.py
```diff
@@ -18,13 +18,24 @@
 from .. import world as w
 
 # Hard-coded workflow-server URL override for testing.
-# Set this to test against a different workflow-server version.
-# Leave empty string for production (uses default vercel-workflow.com).
+# Set this to test against a different workflow-server version, or leave it
+# empty and set VERCEL_WORKFLOW_SERVER_URL in the environment instead.
+# Empty and unset means production (uses default vercel-workflow.com).
```
msullivan: Maybe let's drop the hardcoded override environment variable?

msullivan (approved):

src/vercel/_internal/workflow/worlds/vercel.py
fantix: Sure! I almost dropped it during the first commit but decided otherwise seeing the JS repo kept it. Fixed now!

PR #250, opened 2026-08-06
workflow: don't call Future.set_result without checking for cancellation

msullivan:
If a task waiting on one of these hooks was cancelled, then the future
gets cancelled and the set_result call will blow up.

This bug has been around forever but was made quite a bit worse by
it would get blown up, but it was fine because it had already
registered its follow-up call. There would be a traceback but life
would go on. Now an idle_hook failure actually blows up the loop.

We *could* make it so that we swallow idle_hook failures, but I think
it is actually probably better to let them fail, since they *are* bugs
that we should fix, and most of them won't be as benign as this one
previously was.

### DIFF
--- a/changes/vercel/238.bugfix.md
+++ b/changes/vercel/238.bugfix.md
@@ -1,0 +1,1 @@
+Prevent errors when tasks waiting on steps or hooks are cancelled.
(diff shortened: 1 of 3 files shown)


fantix (approved):
Yeah I think this makes sense

PR #251, opened 2026-08-06
workflow: serve steps on the workflow queue

fantix:
Step invocations used to travel on a dedicated `__wkf_step_*` topic consumed by their own handler. They now ride the workflow topic as a stepId/stepName pair on WorkflowInvokePayload, and workflow_handler dispatches those to step execution instead of replay. stepName has to be on the payload because the merged topic no longer encodes it: the queue name identifies the workflow, not the step.

This matches the TypeScript SDK, which retired its step route in vercel/workflow#3061, and halves the queue triggers deployed for a workflow function -- the Python builder emits one per registered subscription, and a registry now registers one.

Steps are still their own queue hop; the eager inline execution the TypeScript SDK gained alongside this is not ported.

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -525,7 +525,11 @@ async def workflow_handler(
     namespace: str | None = None,
 ) -> w.QueueContinuation | None:
     world = w.get_world()
-    run_id = w.WorkflowInvokePayload.model_validate(message).run_id
+    req = w.WorkflowInvokePayload.model_validate(message)
+    if req.step_id is not None:
+        return await _execute_step(req, queue_name=queue_name, registry=registry)
+
+    run_id = req.run_id
     workflow_run = await world.runs_get(run_id)
     if workflow_run.status == "pending":
         try:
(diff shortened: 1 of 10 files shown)


src/vercel/_internal/workflow/runtime.py
```diff
@@ -525,7 +525,11 @@ async def workflow_handler(
     namespace: str | None = None,
 ) -> w.QueueContinuation | None:
     world = w.get_world()
-    run_id = w.WorkflowInvokePayload.model_validate(message).run_id
+    req = w.WorkflowInvokePayload.model_validate(message)
+    if req.step_id is not None:
+        return await _execute_step(req, queue_name=queue_name, registry=registry)
```
msullivan, line 530: Are these the semantics that match what TS does?
Run the step without invoking the workflow after?

msullivan (approved):

src/vercel/_internal/workflow/runtime.py
fantix: Yeah, the workflow invocation is decided inside `_execute_step()` - which matches what TS does.

PR #255, opened 2026-08-06
Ignore environment-specific uv locks

scotttrinh:
The issue we are resolving is that local development is required to use the company-mandated proxy PyPI registry, but that registry and the resolution it produces do not translate to contributors or CI environments outside of our local machines. A checked-in `uv.lock` therefore captures one machine's registry-specific view of the dependency graph and makes it repository state for everyone else. This change keeps the lockfile as a useful local artifact while stopping it from being committed.

`uv` remains the resolver. On a fresh checkout, contributors run `uv sync` or `uv lock` to generate the local lockfile, and the documentation now explains that the file is environment- and registry-specific. The normal CI workflow no longer passes `--locked` to its initial `uv sync`, so each runner can generate its own lockfile before running locked tools. The latest-dependencies workflow continues to use `--upgrade`. The news-fragment workflow explicitly bootstraps with `uv lock` and then runs its existing locked check. `scripts/workspace.py` still uses locked workspace metadata by default, but now tells callers to create the local lockfile first. Release and build tooling continue to use the generated lockfile: workspace ordering and package metadata are resolved through `uv`, and bundle vendoring still reads the generated lockfile to obtain the exact versions it must vendor.

One local build hiccup was the nested vendoring path. The existing bundle builder runs `uvx` to provide the vendoring tool, and that tool invokes `pip install` internally. The company registry does not work for pip until the local credentials are available through netrc, so the first bundle attempt returned 401 errors even though top-level uv resolution worked. After the credentials were patched into netrc, the nested pip installs, license fetches, bundle builds, and distribution verification all completed successfully. The final build produced and installed all 12 regular wheels; the bundle build also completed using the documented shared-vendor version override when the direct public PyPI metadata lookup was unavailable.

The dependency audit covered the root project, all 12 workspace packages, and all 9 examples. Runtime declarations, optional extras, build requirements, workspace bounds, source imports, example references, and built wheel metadata were checked. The root development floors now include FastAPI 0.141.1, Hatchling 1.31.0, Hypothesis 6.165.0, pytest 9.1.1, pytest-asyncio 1.4.0, tox 4.58.0, tox-uv-bare 1.36.0, uvicorn 0.52.1, and zizmor 1.29.0, with the existing major-version bounds preserved and the tox requirements capped consistently. The newer zizmor version also reported the intentionally pinned workflow CLI install, so that existing step has a narrow rule suppression.

## CI consistency follow-up

Following review feedback, the regular CI matrix now has one bootstrap job that runs `uv lock` once and uploads the generated file as an auditable workflow artifact. Every Python-version job depends on that job, downloads the same lockfile, and keeps `uv sync --locked`, so the matrix cannot drift between independent resolutions. The lockfile is downloaded before setup-uv runs, which lets setup-uv key its reusable dependency cache from that exact lockfile. The cache remains separate from the auditable artifact: it accelerates later runs without turning platform-specific downloaded packages into repository or workflow-resolution state.


.github/workflows/ci.yml
```diff
@@ -75,12 +75,13 @@ jobs:
...
       # the read-only `inspect` path does not use.
       - name: Install the workflow CLI
+        # zizmor: ignore[adhoc-packages]
         run: >-
           npm install workflow@5.0.0-beta.38 --no-save --no-audit --no-fund
           --ignore-scripts --prefix "$RUNNER_TEMP/workflow-cli"
 
       - name: Sync dependencies
-        run: uv sync --all-packages --locked --python ${{ matrix.python-version }}
+        run: uv sync --all-packages --python ${{ matrix.python-version }}
```
fantix: Let's keep the `--locked` and extract a separate `Bootstrap lockfile` job that does only `uv lock` + `upload-artifact`, that each test job depends on and downloads artifact from, for the purpose of 1) consistent test environment across Python versions, and 2) auditable artifact that can be manually downloaded and reproduced.

fantix (approved):

.github/workflows/ci.yml
scotttrinh: Yep, good call. I also added caching to the setup-uv action to hopefully cut down on runtimes, lemme know if that makes sense!

fantix (approved):

.github/workflows/ci.yml
fantix: Looks great!

elprans (changes requested):
Let's hold off with merging this.  Re-generating `uv.lock` defeats all the benefits of having said lock.   I think we can fix this by making `uv lock` (and `uv add`) reference upstream PyPI directly instead.

scotttrinh:
> I think we can fix this by making uv lock (and uv add) reference upstream PyPI directly instead.

Doesn't that defeat the purpose of our proxy then?

scotttrinh:
No longer needed.

PR #256, opened 2026-08-06
workflow: streaming (1/2)

fantix:
The write side of workflow streaming. A step pushes chunks to the run's stream while it runs, and a client reads them without waiting for the run to finish.

```python
@app.step
async def summarize(*, doc: str) -> str:
    writable = get_writable()
    async for token in llm.stream(doc):
        await writable.write(token)
    ...

@app.step
async def done() -> None:
    await get_writable().close()
```

Chunks are readable immediately from the TypeScript SDK (`run.readable`), the dashboard, and `workflow inspect stream`.

## What's here

- `get_writable()` inside a step, with `write()` / `write_from()` / `drain()` / `close()`.
- Length-prefixed frames around the same `devl` payload the event log uses. One write, one frame, one chunk index — that is what makes `start_index + frames_consumed` a correct resume position.
- `World.streams_*`, over HTTP for the Vercel world and one file per chunk for the local one.
- The writer group-commits: `write()` returns once the chunk is buffered, so a producer keeps filling while a request is in flight and the group leaves as one request.
- A `bytes` chunk goes out as a `Uint8Array`, so a TypeScript consumer can pipe the stream straight into a `Response` — that body only accepts `Uint8Array` chunks, and the `ArrayBuffer` a payload would use raises there.

## Behaviour worth knowing

- **Nothing closes a stream implicitly.** It spans steps, and a closed stream cannot be reopened. A run that never closes leaves readers waiting until it expires.
- **A step is not recorded complete until its chunks are durable**, so a write failure fails the step instead of surfacing to a reader that never sees the chunk.
- **Retries re-stream.** A step that fails halfway already wrote what it wrote. Keep chunks idempotent.

## Not here — follows in 2/2

- Reading from Python: `Run.readable()` and the reconnecting reader.
- Passing a stream into a step as an argument.
- Encryption (`encr` / `encp`) and compression.

## Deliberate divergences from `@workflow/core`

- **Stricter step completion.** TS gives the flush 500 ms and finishes the step either way. Here it always waits, which is safe because `drain()` waits only on chunks already handed to it, so it cannot hang.
- **A read starting past the end of a *closed* stream returns empty.** On an open stream both SDKs wait, which is right — the chunk may still be written. Once the stream is closed it never will be, but `@workflow/world-local` marks every skipped file delivered, EOF marker included, so its poll never sees the end and waits forever.
- **An unknown format prefix raises**, naming the prefix, where TS silently drops the chunk.

## Verified

- 96 unit tests: framing, the writer's bounds and failure modes, both worlds' transports.
- The real `workflow` CLI reads the chunks a Python run wrote — added to the existing interop suite.
- Ran against a deployed project: streaming across a durable sleep, batched writes, close, and resume at both a positive and a negative `startIndex`.

Three commits, one per layer, each reviewable on its own.

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -974,7 +974,7 @@ async def _execute_step(
 
     # Bound before the try so the failure path can flush whatever the step
     # managed to write before it raised.
-    step_streams = _StepStreams(run_id=req.workflow_run_id)
+    step_streams = _StepStreams(run_id=req.run_id)
 
     try:
         if not step_run.started_at:
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 import contextvars
 import dataclasses
 import functools
@@ -10,7 +11,7 @@ import re
 import sys
 import traceback
 from collections import deque
-from collections.abc import Callable, Coroutine
+from collections.abc import AsyncIterator, Callable, Coroutine
 from datetime import datetime, timedelta
 from typing import Any, Generic, Literal, ParamSpec, TypeVar
 
@@ -151,6 +152,45 @@ class _StepStreams:
 
     run_id: str
     writers: dict[str, streams.WorkflowStreamWriter] = dataclasses.field(default_factory=dict)
+    _task_group: anyio.abc.TaskGroup | None = None
+
+    @contextlib.asynccontextmanager
+    async def dispatching(self) -> AsyncIterator[None]:
+        """Own the writers' background sends for the duration of the block.
+
+        A writer sends from a task rather than from `write()`, so those tasks
+        need a group whose lifetime covers every write *and* the flush that
+        follows. Leaving the block waits for them, so nothing is still in
+        flight once it returns.
+
+        If the block raises, the chunks the step did manage to stream are
+        flushed first: those are as real as any other, and a reader tailing the
+        run should see the progress that led up to the failure. Best-effort by
+        construction -- the step is already failing, and a flush error here
+        would replace the cause with a symptom. (`@workflow/core` does not do
+        this at all: its `ops` flush lives only on the success path, so a
+        throwing step's buffered chunks are never awaited.)
+
+        The error comes back out as itself rather than wrapped in the task
+        group's exception group, because callers up the stack switch on
+        `FatalError` and count retries.
+        """
+        error: BaseException | None = None
+        async with anyio.create_task_group() as task_group:
+            self._task_group = task_group
+            try:
+                yield
+            except Exception as exc:
+                error = exc
+                await self.drain_quietly()
+            except BaseException as exc:
+                # Cancellation, or the interpreter going down. Draining would
+                # need a shield and could stall the teardown it is racing.
+                error = exc
+            finally:
+                self._task_group = None
+        if error is not None:
+            raise error
 
     def writer(
         self, namespace: str | None, *, reentrant_ctx_on_err: bool = True
@@ -164,6 +204,8 @@ class _StepStreams:
         scramble the order the caller wrote in. Sharing one serial sink makes
         that pattern correct instead of subtly wrong.
         """
+        if self._task_group is None:
+            raise RuntimeError("stream writers are only available while a step is running")
         name = streams.workflow_run_stream_id(self.run_id, namespace)
         writer = self.writers.get(name)
         if writer is None:
@@ -171,6 +213,7 @@ class _StepStreams:
                 world=w.get_world(),
                 run_id=self.run_id,
                 name=name,
+                task_group=self._task_group,
                 reentrant_ctx_on_err=reentrant_ctx_on_err,
             )
             self.writers[name] = writer
@@ -1007,27 +1050,30 @@ async def _execute_step(
             )
         )
         streams_token = _step_streams_ctx.set(step_streams)
-        # Execute the step function
-        try:
-            result = await step.func(*args, **kwargs)
-        finally:
-            _step_ctx.reset(token)
-            _step_streams_ctx.reset(streams_token)
+        async with step_streams.dispatching():
+            # Execute the step function
+            try:
+                result = await step.func(*args, **kwargs)
+            finally:
+                _step_ctx.reset(token)
+                _step_streams_ctx.reset(streams_token)
 
-        # A stream write returns as soon as it is buffered, so the chunks this
-        # step wrote are not durable yet. Force them out before recording the
-        # step as complete, so "the step finished" keeps implying "everything it
-        # streamed is readable" -- and so a failure to write fails the step
-        # rather than being discovered by a reader that never sees the chunk.
-        #
-        # Stricter than `@workflow/core`, deliberately. Its step executor caps
-        # the same flush at 500ms and completes the step regardless, handing the
-        # rest to `waitUntil`, because its `ops` include lock-release polling on
-        # a `WritableStream` the user may hold open across steps -- that can
-        # never settle, so it cannot be awaited unbounded. Nothing here waits on
-        # a lock: `drain()` waits only for chunks already handed over, so it
-        # always terminates and needs no escape hatch.
-        await step_streams.drain()
+            # A stream write returns as soon as it is buffered, so the chunks
+            # this step wrote are not durable yet. Force them out before
+            # recording the step as complete, so "the step finished" keeps
+            # implying "everything it streamed is readable" -- and so a failure
+            # to write fails the step rather than being discovered by a reader
+            # that never sees the chunk.
+            #
+            # Stricter than `@workflow/core`, deliberately. Its step executor
+            # caps the same flush at 500ms and completes the step regardless,
+            # handing the rest to `waitUntil`, because its `ops` include
+            # lock-release polling on a `WritableStream` the user may hold open
+            # across steps -- that can never settle, so it cannot be awaited
+            # unbounded. Nothing here waits on a lock: `drain()` waits only for
+            # chunks already handed over, so it always terminates and needs no
+            # escape hatch.
+            await step_streams.drain()
 
         # Serialize the result
         output = ser.dehydrate(result)
@@ -1039,16 +1085,6 @@ async def _execute_step(
         )
 
     except Exception as e:
-        # Flush what the step did stream before failing: those chunks are as
-        # real as any other, and a reader tailing the run should see the
-        # progress that led up to the failure. Best-effort by construction --
-        # the step is already failing, and a drain error here would replace the
-        # cause with a symptom.
-        #
-        # `@workflow/core` does not do this: its `ops` flush lives only on the
-        # success path, so a throwing step's buffered chunks are never awaited.
-        await step_streams.drain_quietly()
-
         # step.attempt was incremented by step_started
         current_attempt = step_run.attempt
         error_text = "".join(traceback.format_exception_only(type(e), e)).strip()
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -152,7 +152,9 @@ class _StepStreams:
     run_id: str
     writers: dict[str, streams.WorkflowStreamWriter] = dataclasses.field(default_factory=dict)
 
-    def writer(self, namespace: str | None) -> streams.WorkflowStreamWriter:
+    def writer(
+        self, namespace: str | None, *, reentrant_ctx_on_err: bool = True
+    ) -> streams.WorkflowStreamWriter:
         """The writer for this run's *namespace* stream, created on first use.
 
         One writer per stream per step, deliberately. Handing out a fresh writer
@@ -165,7 +167,12 @@ class _StepStreams:
         name = streams.workflow_run_stream_id(self.run_id, namespace)
         writer = self.writers.get(name)
         if writer is None:
-            writer = streams.WorkflowStreamWriter(w.get_world(), self.run_id, name)
+            writer = streams.WorkflowStreamWriter(
+                world=w.get_world(),
+                run_id=self.run_id,
+                name=name,
+                reentrant_ctx_on_err=reentrant_ctx_on_err,
+            )
             self.writers[name] = writer
         return writer
 
@@ -202,7 +209,11 @@ def get_step_metadata() -> StepInfo:
         raise RuntimeError("get_step_metadata() can only be called inside a step") from None
 
 
-def get_writable(*, namespace: str | None = None) -> streams.WorkflowStreamWriter:
+def get_writable(
+    *,
+    namespace: str | None = None,
+    reentrant_ctx_on_err: bool = True,
+) -> streams.WorkflowStreamWriter:
     """The run's writable stream, for streaming output as the step runs.
 
     Chunks become readable immediately -- through ``run.readable`` on the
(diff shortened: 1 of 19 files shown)


msullivan:
> Retries re-stream. A step that fails halfway already wrote what it wrote. Keep chunks idempotent.

I think that streams *ought* to have a way to have an idempotency key, but we'll need to raise this upstream I think.
In temporal, workflows themselves can write to streams, and so they must be doing some idempotency thing for it. In workflow, the workflow needs to have a step do the right, but the step might get retried

fantix:
> > Retries re-stream. A step that fails halfway already wrote what it wrote. Keep chunks idempotent.
> 
> I think that streams _ought_ to have a way to have an idempotency key, but we'll need to raise this upstream I think. In temporal, workflows themselves can write to streams, and so they must be doing some idempotency thing for it. In workflow, the workflow needs to have a step do the right, but the step might get retried

I think that's valid; recorded as vercel/workflow#3387

src/vercel/_internal/workflow/runtime.py
```diff
@@ -152,6 +202,35 @@ def get_step_metadata() -> StepInfo:
...
+    waiting for the step or the run to finish. This SDK has no reader of its own
+    yet; :meth:`vercel._internal.workflow.world.World.streams_get` is the
+    unstable way in.
+
+    Pass *namespace* to write to a second, independent stream on the same run.
+
+    Must be called from within a step body; raises ``RuntimeError`` otherwise.
+    A workflow body cannot stream: unlike the TypeScript SDK it cannot even take
+    a handle to pass into a step, which is why a streaming step calls this
+    itself.
```
msullivan, line 273: What's the rationale for not supporting the handle passing that TypeScript does? Is it just that they were fitting it into a well-established stream framework in JS-land that isn't applicable here?

We might eventually want to support writing to the stream from external clients, right? I think TS does?

src/vercel/_internal/workflow/runtime.py
```diff
@@ -892,6 +992,13 @@ async def step_handler(
         # TODO: Check if this is a fatal error (would need FatalError class)
         # For now, treat all errors as potentially retryable
 
+        # Flush what the step did stream before failing: those chunks are as
+        # real as any other, and a reader tailing the run should see the
+        # progress that led up to the failure. Best-effort by construction --
+        # the step is already failing, and a drain error here would replace the
+        # cause with a symptom.
+        await step_streams.drain_quietly()
```
msullivan: Does this match JS?
I could see an argument for going the other direction, but this is probably right.

src/vercel/_internal/workflow/world.py
```diff
@@ -677,6 +677,37 @@ class PaginatedResult(BaseModel, Generic[T]):
...
+    """Whether the stream has been closed."""
+
+
+class StreamChunk(BaseModel):
+    index: int
+    data: bytes
+
+
+class StreamChunksPage(BaseModel):
+    """One page of already-written chunks -- a snapshot, not a live read."""
```
msullivan, line 731: Maybe define page somewhere

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+    if len(payload) > MAX_FRAME_SIZE:
+        raise ser.SerializationError(
+            f"Stream chunk of {len(payload)} bytes exceeds the maximum frame size "
+            f"({MAX_FRAME_SIZE}); split the data into smaller chunks before writing"
+        )
+    return len(payload).to_bytes(FRAME_HEADER_SIZE, "big") + payload
+
+
+def _reduce_uint8array(value: Any) -> Any:
+    """devalue reducer sending a chunk's ``bytes`` as a view. Falsy declines.
```
msullivan: "Falsy declines" is a weird phrasing and also not true.
Instead it's doing `type(value) is not bytes` which is pretty weird!

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+    lead with pipes a run's readable straight into a `Response`, and a body
+    stream takes `Uint8Array` chunks only -- an `ArrayBuffer` chunk raises
+    ``Received non-Uint8Array chunk``. So a step writing ``b"..."`` has to
+    frame it as a view or every consumer doing that breaks.
+
+    Chunk-local on purpose. Python has one bytes type, so whichever form the
+    encoder picks the other becomes a one-way landing; the payload boundary
+    already chose ``ArrayBuffer`` and has tests pinning it. Nothing downstream
+    of a payload cares which form a ``bytes`` *field* took, so there is no
+    reason to move that.
```
msullivan: This is a lot of comment text and honestly I still only half understand what is going on here.
I'm not sure what this distinction with a ``bytes`` *field* is.

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+    if type(value) is not bytes:
+        return False
+    # A `memoryview` rather than the `bytes` itself, and not for speed: devalue
+    # indexes by identity, so returning the same object would make the view
+    # reference its own slot instead of a buffer. A distinct object flattens to
+    # the `["ArrayBuffer", …]` entry the view is supposed to point at -- and a
+    # view over the same memory copies nothing to get one.
+    return memoryview(value)
```
msullivan: hmmmmm

src/vercel/_internal/workflow/streams.py
msullivan: I think it might be better to make `bytes` always be Uint8Array, have ArrayBuffer be `bytearray`?? And the always have the same reducers?

src/vercel/_internal/workflow/streams.py
msullivan: Oh, I now understand what the reducer protocol is, ok. Comment still unhelpful?

Also looking at the reducer protocol, I see that it uses `_is_js_truthy` for checking whether a reducer fired, which *really* seems like it ought not be necessary for our Python implementation??

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+                f"it was truncated mid-frame"
+            )
+
+
+# ═══════════════════════════════════════════════════════════════════════════
+# writer
+# ═══════════════════════════════════════════════════════════════════════════
+#
+# Ported from `@workflow/core`'s `WorkflowServerWritableStream`, whose shape is
+# not obvious and is worth restating.
```
msullivan: slop throat clearing comment

msullivan: (the overall comment here is useful, though, but "not obvious and is worth restating"...)

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+
+MAX_CHUNKS_PER_BATCH = 1000
+"""Chunks in one write request, matching the server's per-batch cap."""
+
+MAX_BYTES_PER_BATCH = 1024 * 1024
+"""Cumulative bytes in one write request, under platform body limits."""
+
+
+class WorkflowStreamWriter:
+    """A group-committing writer for one ``(run_id, name)`` stream.
```
msullivan: hmmm

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+        """
+        await self._enqueue(encode_value(value))
+
+    async def write_from(self, source: AsyncIterable[Any]) -> None:
+        """Append every item *source* yields, in order.
+
+        The counterpart of piping a stream into this one: a step forwarding an
+        upstream async iterator (LLM tokens, an HTTP body) wants this rather
+        than a hand-rolled loop, and it keeps the group-commit buffer fed
+        without the caller thinking about batching.
```
msullivan: it's identical to a hand-rolled loop, though?

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+                        return
+                    group, octets = self._take_group()
+                    self._inflight_chunks = len(group)
+                    self._inflight_bytes = octets
+
+                try:
+                    await self._send(group)
+                except BaseException as error:
+                    cancelled = isinstance(error, asyncio.CancelledError)
+                    async with self._condition:
```
msullivan: I was potentially nervous about the case where this blocks and then gets cancelled again, but this currently won't ever block because nothing else does a real `await` on anything while holding the condition.

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+                        # The group did not land, so it goes back at the head
+                        # either way: the buffer holds everything not known to
+                        # be durable, and dropping it here would lose chunks
+                        # whose `write()` had already returned.
+                        self._buffer[:0] = group
+                        self._buffered_bytes += octets
+                        self._inflight_chunks = 0
+                        self._inflight_bytes = 0
+                        self._dispatch = None
```
msullivan: Maybe have a method for this?

src/vercel/_internal/workflow/streams.py
```diff
@@ -0,0 +1,435 @@
...
+            return
+        await self.drain()
+        await self._world.streams_close(self._run_id, self._name)
+        self._closed = True
+
+    async def __aenter__(self) -> WorkflowStreamWriter:
+        return self
+
+    async def __aexit__(self, exc_type: object, exc: object, tb: object) -> None:
+        if exc_type is None:
```
msullivan: No close at all on exception?

src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -168,6 +230,24 @@ def write_json(path: pathlib.Path, data: w.BaseModel | dict, *, overwrite: bool
...
 
+class UnsafeEntityIdError(ValueError):
+    """An id that would not stay put inside the data directory."""
+
+    def __init__(self, kind: str, value: str) -> None:
+        super().__init__(f"unsafe {kind}: {value!r}")
+
+
+def assert_safe_entity_id(kind: str, value: str) -> None:
+    """Reject an id that cannot be used as a path segment.
```
msullivan, line 281: Huh I thought these had gotten base64d already?

src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -846,6 +929,189 @@ def _events_create_impl(self, run_id: str | None, data: w.Event) -> w.EventResul
...
+    def _append_chunk(
+        self, run_id: str, name: str, chunk_id: str, payload: bytes, *, eof: bool
+    ) -> None:
+        self._register_stream(run_id, name)
+        path = self._chunk_dir(name) / f"{chunk_id}.bin"
+        path.parent.mkdir(parents=True, exist_ok=True)
+        atomic_write(path, bytes([_EOF_MARKER if eof else 0]) + payload)
+
+    async def streams_write(self, run_id: str, name: str, chunk: bytes) -> None:
+        chunk_id = f"chnk_{self.monotonic_ulid(None)}"
```
msullivan: Have a helper that generates the chunk ids instead of repeating all these times

msullivan: I assume that the chunk id is forced by compatibility, which is why we can't just use integers?
(And probably I'm missing some other reason we can't just use integers?)

src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -41,6 +42,67 @@ def is_step_terminal(status: str) -> bool:
...
+    """Just a chunk's EOF flag, without reading its payload."""
+    try:
+        with path.open("rb") as f:
+            return f.read(1) == bytes([_EOF_MARKER])
+    except FileNotFoundError:
+        return False
+
+
+def _encode_chunks_cursor(index: int) -> str:
+    return base64.b64encode(json.dumps({"i": index}).encode()).decode("ascii")
```
msullivan, line 127: I assume this is for TS format compatability somehow? How exactly?
I'm kind of missing something I think with these opaque cursors?

src/vercel/_internal/workflow/worlds/local.py
```diff
@@ -846,6 +929,189 @@ def _events_create_impl(self, run_id: str | None, data: w.Event) -> w.EventResul
...
+            await asyncio.sleep(_POLL_INTERVAL_SECONDS)
+            for path in self._chunk_files(name):
+                if path.name in delivered:
+                    continue
+                delivered.add(path.name)
+                eof, payload = _read_chunk(path)
+                if eof:
+                    return
+                if payload:
+                    yield payload
```
msullivan, line 1105: I think this ought to be implementable by a loop that reuses `stream_get_chunks`? (Or rather, maybe a shared helper that just uses an int position index)

msullivan (approved):

src/vercel/_internal/workflow/runtime.py
fantix: The next 2/2 PR added passing handles! This comment should've mentioned the phased impl tho.

> writing to the stream from external clients

and let the workflow consume such external data? That sounds like what hooks can do. I didn't see TS doing external streaming, but maybe yes?

fantix: Ohh if it's just for external processes/tools called by a step to be able to reply to the same stream, instead of routing back through the step, then yes that makes sense to me.

src/vercel/_internal/workflow/runtime.py
fantix: Right, this is not a 100% JS match. I think JS is just designed in a way that waiting on the failure-path is not convenient to implement (and maybe not necessary due to retries?)

src/vercel/_internal/workflow/world.py
fantix: Fixed!

src/vercel/_internal/workflow/streams.py
fantix: Yeah, this is now dropped after landing #272

src/vercel/_internal/workflow/streams.py
fantix: Done as suggested in #272

src/vercel/_internal/workflow/streams.py
fantix: Reworded!

src/vercel/_internal/workflow/streams.py
fantix: Yes! Reworded manually.

src/vercel/_internal/workflow/streams.py
fantix: Right, added a comment on `self._condition.`

src/vercel/_internal/workflow/streams.py
fantix: Fixed

src/vercel/_internal/workflow/streams.py
fantix: This was on purpose so that a retried step could re-enter the same stream. I'll add an option to control this behavior, as well as to make it more explicit.

src/vercel/_internal/workflow/worlds/local.py
fantix: Yeah, this double guard is a copy of the JS local world, for tests/CLI direct calls I think.

src/vercel/_internal/workflow/worlds/local.py
fantix: Helper added. Re integers: yeah compatibility and that concurrent steps writing to the same stream shouldn't produce conflicting chunk IDs.

src/vercel/_internal/workflow/worlds/local.py
fantix: Actually it doesn't have to be compatible. Added a doc string.

src/vercel/_internal/workflow/worlds/local.py
fantix: Right! Added a shared helper `_scan_chunks()`.

fantix:
Also added a commit to use anyio instead of raw asyncio `Task`.

PR #258, opened 2026-08-06
workflow: streaming (2/2)

fantix:
Stacks on #256 — review that first; this branch adds two commits on top.

The read side, plus letting a workflow hand its stream to the steps that write.

```python
run = await start(analyze, document=text)

async for chunk in run.readable():
    print(chunk)
```

## What's here

- `run.readable()` yields the values a run's steps stream, in write order, ending when a step closes the stream. `readable_bytes()` narrows that to `bytes` for an HTTP body.
- `run.stream_info()` reports the tail index and whether the stream is closed; `run.list_streams()` names every stream the run has; `read_stream(run_id, name)` reads one by name.
- `get_writable()` now answers in a workflow body too, so one workflow can name a stream and pass it to every step that writes to it:

```python
@app.step
async def summarize(*, document: str, out: WorkflowWritable) -> str:
    await out.write("starting")
    ...

@app.workflow
async def analyze(*, document: str) -> str:
    out = get_writable()
    return await summarize(document=document, out=out)
```

## Resuming

`start_index` picks where to start. A positive index resumes exactly — hand back the index you last saw and pass it in next time. A negative one reads that many chunks back from the end, and is single-shot: it resolves against wherever the tail was at connect time, so reopening would shift the window rather than resume it.

A read reconnects on its own when the transport drops, which it will — the server ends a long read at its own time limit. One frame is one stored chunk, so the resume position is exact and nothing is duplicated or skipped.

## Deliberate divergences from `@workflow/core`

- **`run.readable()` is a method, not a property.** Each call opens its own read. A property would make iterating it twice quietly start a second one; TS gets away with it because a `ReadableStream` can be handed out repeatedly, where a Python async generator is single-use.
- **The workflow body's writable raises instead of failing obscurely.** TS returns an object carrying `WritableStream`'s prototype, so writing to it throws a brand-check error. Here both contexts share a `WorkflowWritable` interface and the workflow-side one raises with what to do instead.

## Notes for review

- `get_writable()` now returns `WorkflowWritable` rather than `WorkflowStreamWriter`. Source-compatible — the writer implements it.
- Reviving a stream passed into a step registers it with that step, so it drains before the step is recorded complete. That required installing the step context before the input is hydrated, not just around the body.
- No new changelog fragment: this updates #256's, since the two do not release separately.

## Verified

- 24 new unit tests, mostly on the reconnect: the resume index after a break, a frame cut in half by one arriving whole, a clean EOF not reconnecting, and both caps — including a world that ignores `start_index`, which resets the consecutive cap forever and can only be stopped by the absolute one.
- The existing CLI interop suite still passes, so the JS reader agrees with what Python writes.

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -151,7 +151,9 @@ class _StepStreams:
     """
 
     run_id: str
-    writers: dict[str, streams.WorkflowStreamWriter] = dataclasses.field(default_factory=dict)
+    writers: dict[tuple[str, str], streams.WorkflowStreamWriter] = dataclasses.field(
+        default_factory=dict
+    )
     _task_group: anyio.abc.TaskGroup | None = None
 
     @contextlib.asynccontextmanager
@@ -162,6 +164,10 @@ class _StepStreams:
         need a group whose lifetime covers every write *and* the flush that
         follows. Leaving the block waits for them, so nothing is still in
         flight once it returns.
+
+        It has to open before the step's input is hydrated, not just around the
+        body: a stream the workflow passed in becomes a writer while the input
+        is being read, and a writer with no group behind it cannot send.
 
         If the block raises, the chunks the step did manage to stream are
         flushed first: those are as real as any other, and a reader tailing the
@@ -195,28 +201,40 @@ class _StepStreams:
     def writer(
         self, namespace: str | None, *, reentrant_ctx_on_err: bool = True
     ) -> streams.WorkflowStreamWriter:
-        """The writer for this run's *namespace* stream, created on first use.
+        """The writer for this run's *namespace* stream."""
+        return self.writer_for(
+            self.run_id,
+            streams.workflow_run_stream_id(self.run_id, namespace),
+            reentrant_ctx_on_err=reentrant_ctx_on_err,
+        )
+
+    def writer_for(
+        self, run_id: str, name: str, *, reentrant_ctx_on_err: bool = True
+    ) -> streams.WorkflowStreamWriter:
+        """The writer for one stream, created on first use.
 
         One writer per stream per step, deliberately. Handing out a fresh writer
         each call would give each its own buffer over the same stream, and two
         buffers flushing independently interleave their chunks by whichever
         request happens to win -- so `get_writable()` twice in a loop would
         scramble the order the caller wrote in. Sharing one serial sink makes
-        that pattern correct instead of subtly wrong.
+        that pattern correct instead of subtly wrong, and it is why a handle the
+        workflow passed in resolves to the same writer the step would have got
+        by asking for the stream itself.
         """
         if self._task_group is None:
             raise RuntimeError("stream writers are only available while a step is running")
-        name = streams.workflow_run_stream_id(self.run_id, namespace)
-        writer = self.writers.get(name)
+        key = (run_id, name)
+        writer = self.writers.get(key)
         if writer is None:
             writer = streams.WorkflowStreamWriter(
                 world=w.get_world(),
-                run_id=self.run_id,
+                run_id=run_id,
                 name=name,
                 task_group=self._task_group,
                 reentrant_ctx_on_err=reentrant_ctx_on_err,
             )
-            self.writers[name] = writer
+            self.writers[key] = writer
         return writer
 
     async def drain(self) -> None:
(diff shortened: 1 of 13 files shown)


src/vercel/_internal/workflow/runtime.py
```diff
@@ -1200,6 +1263,95 @@ async def return_value(self) -> Any:
...
+
+    async def list_streams(self) -> list[str]:
+        """Every stream this run has written to, namespaced ones included."""
+        return await self._world.streams_list(self._run_id)
+
+
+def read_stream(
+    run_id: str, name: str, *, start_index: int | None = None
+) -> AsyncGenerator[Any, None]:
+    """Read any stream of a run by name, for a stream :class:`Run` cannot name.
```
msullivan: I'm not sure what "for a stream Run cannot name" means here

src/vercel/_internal/workflow/streams.py
```diff
@@ -173,6 +176,91 @@ def finish(self) -> None:
             )
 
 
+# ═══════════════════════════════════════════════════════════════════════════
+# reader
+# ═══════════════════════════════════════════════════════════════════════════
+
+MAX_RECONNECTS = 50
```
msullivan, line 158: Huh, higher than I'd guess! Matches JS?

src/vercel/_internal/workflow/streams.py
```diff
@@ -173,6 +176,91 @@ def finish(self) -> None:
...
+"""
+
+MAX_TOTAL_RECONNECTS = 1000
+"""Absolute backstop, independent of progress.
+
+The consecutive cap resets on forward progress, which is correct against a
+backend that honours ``start_index``. One that ignored it would re-deliver
+earlier frames, look like progress every time, and never trip that cap -- so
+this guarantees the loop terminates regardless.
+"""
```
msullivan, line 174: I don't totally follow this?

src/vercel/_internal/workflow/streams.py
```diff
@@ -173,6 +176,91 @@ def finish(self) -> None:
...
+    where the tail was at connect time, and re-resolving it after a break would
+    silently move the window. Such a read is single-shot.
+    """
+    resumable = start_index is None or start_index >= 0
+    position = start_index or 0
+    consumed = 0
+    consecutive = 0
+    total = 0
+    max_reconnects = _env_int("WORKFLOW_FRAMED_STREAM_MAX_RECONNECTS", MAX_RECONNECTS)
+    max_total = _env_int("WORKFLOW_FRAMED_STREAM_MAX_TOTAL_RECONNECTS", MAX_TOTAL_RECONNECTS)
```
msullivan, line 203: I think this applied to 1/2 also, but: should these be globals instead of locals?

src/vercel/_internal/workflow/streams.py
```diff
@@ -208,7 +296,97 @@ def finish(self) -> None:
 """Cumulative bytes in one write request, under platform body limits."""
 
 
-class WorkflowStreamWriter:
+class WorkflowWritable(abc.ABC):
+    """One of a run's streams, as :func:`vercel.workflow.get_writable` hands it out.
+
+    Two things wear this interface. In a step it is a
```
msullivan: "wear"

src/vercel/_internal/workflow/streams.py
```diff
@@ -208,7 +296,97 @@ def finish(self) -> None:
...
+        """Wait until every accepted chunk is durably written."""
+
+    @abc.abstractmethod
+    async def close(self) -> None:
+        """Mark the stream complete, ending its readers."""
+
+
+@dataclasses.dataclass(frozen=True)
+class WorkflowStreamHandle(WorkflowWritable):
+    """A stream a workflow body can name but not write to.
```
msullivan: I'd say reference but not name? Why does claude love abusing the world name so much.

msullivan (approved):

src/vercel/_internal/workflow/streams.py
fantix: Yes this is an exact match.

src/vercel/_internal/workflow/streams.py
fantix: Reworded. I think it was just trying to say that, this is a failsafe to end the loop for an ill-implemented world that ignored `start_index`.

src/vercel/_internal/workflow/streams.py
fantix: Ah I believe this makes tests easier to manipulate, and mirrors the JS impl 1:1.

PR #261, opened 2026-08-07
Move typeutils into vercel-internal-core

msullivan:
The annotation predicates and forward-reference resolution were private to
vercel-queue, but they are not queue-specific, and the workflow codec wants
the same reading of a signature.

vercel-queue gains a dependency on vercel-internal-core, and ruff needs
namespace-packages set to see that vercel._internal.core is shared
internals rather than another project's privates.


scotttrinh (approved):
Changes look good, looks like there are some import ordering failures.

PR #262, opened 2026-08-07
queue: let accept_and_handle pick the transport

fantix:
The receive codec is inferred from the subscriber's payload annotation, which
has no way to name a format such as CBOR — `payload_transport_kind()` maps
every structured annotation onto `RawJsonTransport`, and an untyped handler
gets it unconditionally. A caller that already knows the wire format of a
delivery had nowhere to say so: the `transport` argument existed on the private
`_accept_and_handle`, but not on the public entry points.

Adds `transport` to `accept_and_handle` on the async client, the sync client,
and both module-level functions, and exports the `Transport` protocol the
parameter is typed with.

The codec is all it replaces — payload validation is separate
(`InvocationPlan.payload_adapter` runs `validate_python` on whatever the
transport returned), so a typed subscriber is still validated. There is a test
for that.

No behaviour change when the argument is omitted.

First of three:

1. **this PR** — `accept_and_handle(transport=...)`
2. `subscribe(transport=...)`, so a delivery dispatched by a client the SDK did
   not construct also decodes correctly
3. the Vercel workflow world's CBOR transport, which is what needs both

### DIFF
--- a/changes/vercel-queue/accept-and-handle-transport.feature.md
+++ b/changes/vercel-queue/accept-and-handle-transport.feature.md
@@ -1,0 +1,4 @@
+`accept_and_handle()` now takes a `transport`, so a caller that knows the wire
+format of a delivery can supply it instead of relying on the codec inferred
+from the subscriber's payload annotation — which cannot express formats such as
+CBOR. The `Transport` protocol is exported to go with it.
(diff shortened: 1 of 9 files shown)


changes/vercel-queue/accept-and-handle-transport.feature.md
```diff
@@ -0,0 +1,4 @@
+`accept_and_handle()` now takes a `transport`, so a caller that knows the wire
```
elprans, line 1: The intended mechanism for overriding of the transport is to use `Topic[T]` in the annotation instead of `T` and then pass the transport as a kwarg in the topic constructor, (`Topic[T]("topic", transport=special)`)

PR #263, opened 2026-08-07
queue: allow transport override in subscribe

fantix:
This allows further transport override on `vqs.subscribe` that is needed by workflow.

### DIFF
--- a/changes/vercel-queue/subscribe-transport.feature.md
+++ b/changes/vercel-queue/subscribe-transport.feature.md
@@ -1,0 +1,1 @@
+`subscribe()` now takes a `transport` to override the inferred wire codec.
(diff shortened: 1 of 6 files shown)


elprans (changes requested):
Transport override is supposed to be done via `Topic`, why are we adding a separate override?

fantix:
> Transport override is supposed to be done via `Topic`, why are we adding a separate override?

`Topic` does not allow wildcard topics like `__wkf_*`, so I supposed it was for transport override in *[send](https://github.com/vercel/vercel-py/pull/265/changes#diff-df8c76496f7d085cd24954aaafdef98307bac2ba2062334515709fd124fecb5dR299)* only. But yeah, I can also fix `Topic` so that it carries wildcard topics for the rx end too.

PR #266, opened 2026-08-07
queue: add TopicPattern to carry a codec

fantix:
Closes #262
Closes #263

Two gaps, both about naming the wire codec for a subscription.

A `Topic` could already hold a transport, but only the send and poll paths read
it. A push delivery's codec was always the kind inferred from the subscriber's
payload annotation — one of five fixed kinds (json, byte/text buffer,
byte/text stream) — so a concrete `Topic(transport=...)` was silently ignored
on push. And a wildcard subscriber could not use a `Topic` at all, because
`Topic` rejected `*`.

## What changed

- A new `TopicPattern` names `"*"` or a trailing-`*` prefix. It is a sibling of
  `Topic`, not a subclass, so `send()` and `poll()` — which take a `Topic` —
  reject a pattern statically. Its `name` stays a plain `str`, since
  `SanitizedName` means "safe in a request path".
- A transport on the subscribed `Topic` or `TopicPattern` is recorded on the
  subscription, so push deliveries use it too. It has to live there rather than
  on a client: the client that handles a push is often not one you configured —
  `vercel.queue.asgi_app()` builds its own.

## Using it

```python
from vercel.queue import TopicPattern, subscribe

JOBS = TopicPattern[JobPayload]("jobs*", transport=CborTransport())

@subscribe(topic=JOBS, consumer_group="workers")
async def handle(job: JobPayload) -> None:
    ...
```

Deliveries on any `jobs…` topic are decoded with `CborTransport` and validated
into `JobPayload`. A concrete `Topic` behaves as before, and now honours its
transport on push as well.


src/vercel-queue/vercel/queue/_internal/types.py
```diff
@@ -68,11 +73,22 @@ class Topic(Generic[T]):
...
+
+    A pattern is a plain ``str``: ``SanitizedName`` means "safe to put in a
+    request path", which a pattern is not.
+    """
 
     transport: Transport[Any] | None = None
-    """Optional transport used when sending to or polling this topic."""
+    """Optional transport used when sending to, polling, or subscribing to this topic."""
+
+    is_pattern: bool = False
```
elprans: I'd add a new class like `TopicPattern` for added typesafety as opposed to duck-typing `is_pattern`.

elprans (approved):

PR #271, opened 2026-08-09
[apscheduler] add support for externally managed job stores

nsidnev:
Allow to add other non-`default` job stores to a scheduler that will be treated as "source" stores. Jobs returned by these stores will be checked for execution by a configurable interval (30 seconds by default) or the earliest wake-up before that event


integrations/vercel-apscheduler/vercel/integrations/apscheduler/_backends/cache/__init__.py
```diff
@@ -70,6 +64,14 @@ def validate_configuration(self, scheduler: Any) -> dict[str, Any]:
                 "explicit default store, or select the matching backend via "
                 "VERCEL_APSCHEDULER_BACKEND"
             )
+        for alias, store in stores.items():
+            if alias != "default" and isinstance(store, (CacheJobStore, RedisJobStore)):
```
elprans, line 68: I don't understand this requirement. Why is `default` special?

ricardo-agz (approved):

PR #273, opened 2026-08-10
Add vercel.functions.metric()

tobiaslins:
## Summary

- add `vercel.functions.metric(name, value, tags=None)`
- add the invocation-context callback used by the Vercel Python runtime
- safely no-op when no compatible runtime invocation is active

## Why

This gives Python Functions the same custom metric helper as `@vercel/functions`, while keeping transport details owned by the runtime. Documentation is intentionally omitted during the soft launch.

This PR pairs with the Python runtime IPC change in https://github.com/vercel/vercel-internal/pull/178.

## Validation

- `uv run --no-sync poe test vercel vercel-cache` (1062 passed, 27 credential-gated tests skipped)
- `uv run --no-sync poe lint vercel vercel-cache`
- `uv run poe typecheck vercel vercel-cache`
- focused metric/context tests (13 passed)

The required `vercel` feature and `vercel-cache` internal news fragments are included. The local `check-news-fragments` command could not run because uv 0.11.2 removed the `workspace metadata --locked` option still used by the repository script.

### DIFF
--- a/src/vercel-cache/vercel/cache/context.py
+++ b/src/vercel-cache/vercel/cache/context.py
@@ -8,9 +8,15 @@ from vercel.headers import get_headers as _get_headers, set_headers as _set_head
 
 from .types import PurgeAPI
 
+_MetricCallback = Callable[
+    [str, int | float, Mapping[str, str] | None],
+    None,
+]
+
 _cv_wait_until: ContextVar[Callable[[Awaitable[object]], None] | None] = ContextVar(
     "vercel_wait_until", default=None
 )
+_cv_metric: ContextVar[_MetricCallback | None] = ContextVar("vercel_metric", default=None)
 _cv_cache: ContextVar[object | None] = ContextVar("vercel_cache", default=None)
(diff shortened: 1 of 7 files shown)


src/vercel-cache/vercel/cache/context.py
```diff
@@ -8,9 +8,15 @@
...
 
+_MetricCallback = Callable[
+    [str, int | float, Mapping[str, str] | None],
+    None,
+]
+
 _cv_wait_until: ContextVar[Callable[[Awaitable[object]], None] | None] = ContextVar(
     "vercel_wait_until", default=None
 )
+_cv_metric: ContextVar[_MetricCallback | None] = ContextVar("vercel_metric", default=None)
```
elprans, line 19: This is an odd location for this.  `vercel/functions` would be more appropriate.  Also, `ContextVar` is probably not the right vehicle because the callback does not change between requests, so it should just be a global.

elprans: Also, I would reverse this.  Instead of the runtime patching the SDK, the SDK should detect the presence of the runtime and call runtime callbacks.

PR #279, opened 2026-08-11
workflow: read encrypted `encr` payloads

### LINKED ISSUE #115: runs_get: ValidationError on encrypted bytes in WorkflowRun.input (no decryption path in SDK)
SDK version: `vercel==0.5.8`

`runs_get` raises `ValidationError` because the workflow service returns what appears to be encrypted bytes (`b'encr\x9aN\xf2…'`) for `WorkflowRun.input` (specifically the `pending` discriminator branch). The schema in `world.py:124` types this field as `list[bytes] | str | None`, with no admission for an encrypted envelope.

The Python SDK has no decryption code — `worlds/vercel.py:33` `_cbor_tag_hook` only unwraps CBOR tag 64, and `_internal/workflow/` contains no symmetric-decryption path. This suggests the workflow service applies an encryption envelope that the JS SDK strips transparently, but the Python SDK does not.

## Log trace

```
[info] HTTP Request: GET https://vercel-workflow.com/api/v2/runs/wrun_…?remoteRefBehavior=resolve "HTTP/1.1 200 OK"
[error] Traceback (most recent call last):
[error] File "/var/task/_vendor/vercel/_internal/workflow/worlds/vercel.py", line 227, in async_handler
    result = await handler(
             ^^^^^^^^^^^^^^
[error] File "/var/task/_vendor/vercel/_internal/workflow/runtime.py", line 270, in workflow_handler
    workflow_run = await world.runs_get(run_id)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error] File "/var/task/_vendor/vercel/_internal/workflow/worlds/vercel.py", line 268, in runs_get
    return await self._cbor_request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[error] File "/var/task/_vendor/vercel/_internal/workflow/worlds/vercel.py", line 134, in _cbor_request
    return schema.validate_python(result)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error] File "/var/task/_vendor/pydantic/type_adapter.py", line 441, in validate_python
    return self.validator.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error] pydantic_core._pydantic_core.ValidationError: 2 validation errors for tagged-union[NonFinalWorkflowRun,NonFinalWorkflowRun,CancelledWorkflowRun,CompletedWorkflowRun,FailedWorkflowRun]
pending.input.list[bytes]
  Input should be a valid list [type=list_type, input_value=b'encr\x9aN\xf2i\xd6\x8e\…x1c\x12j\xbfo\x9d\xf3* ', input_type=bytes]
    For further information visit https://errors.pydantic.dev/2.12/v/list_type
pending.input.str
  Input should be a valid string, unable to parse raw data as a unicode string [type=string_unicode, input_value=b'encr\x9aN\xf2i\xd6\x8e\…x1c\x12j\xbfo\x9d\xf3* ', input_type=bytes]
    For further information visit https://errors.pydantic.dev/2.12/v/string_unicode
```

The `b'encr…'` magic prefix suggests this is an explicit encryption envelope (possibly the same `encr` prefix used elsewhere in the workflow service for symmetric AEAD).

fantix:
This PR adds read support for encrypted workflow payloads with `encr` as the format tag. Writing `encr` is left for later; the payloads this SDK produces stay plain `devl`.

- `serialization.hydrate(..., key=)` now handles `encr`.
- `encryption.py`: HKDF-SHA256 over the deployment secret (32 zero-byte salt, `info = "<projectId>|<runId>"`) and the AES-GCM envelope.
- AES-GCM comes from `cryptography`, in a new `vercel[encryption]` extra. Without the extra, it still works for plaintext payloads and an encrypted one raises an error naming the extra.

`encp` (X25519) is still reported as unsupported.

Fixes #115

### DIFF
--- a/src/vercel/_internal/workflow/runtime.py
+++ b/src/vercel/_internal/workflow/runtime.py
@@ -232,9 +232,18 @@ class WorkflowOrchestratorContext:
     _ctx: contextvars.ContextVar[Self] = contextvars.ContextVar("WorkflowContext")
 
     def __init__(
-        self, events: list[w.Event], *, seed: str, started_at: int, registry: core.Workflows
+        self,
+        events: list[w.Event],
+        *,
+        seed: str,
+        started_at: int,
+        registry: core.Workflows,
+        run_key: bytes | None = None,
     ):
         self.events = events
+        # The key is per-run, so it is resolved once and reused for every
+        # payload in the replay.
+        self.run_key = run_key
         # List of Out-of-order HookReceivedEvent: such events may arrive at any time unexpectedly,
         # so we stash them separately for delayed consumption
         self.ooo_hook_received_events: deque[w.HookReceivedEvent] = deque()
@@ -267,7 +276,9 @@ class WorkflowOrchestratorContext:
                 obj = getattr(obj, attr)
 
             what = f"the input of run {workflow_run.run_id}"
-            args, kwargs = ser.call_arguments(ser.hydrate(workflow_run.input, what=what), what=what)
+            args, kwargs = ser.call_arguments(
+                ser.hydrate(workflow_run.input, what=what, key=self.run_key), what=what
+            )
 
             token = self._ctx.set(self)
             try:
@@ -468,7 +479,11 @@ class WorkflowOrchestratorContext:
                 case w.StepCompletedEvent(event_data=w.StepCompletedEventData(result=data)):
                     sus = self.suspensions.pop(event.correlation_id)
                     assert isinstance(sus, Suspension)
-                    result = ser.hydrate(data, what=f"the result of step {event.correlation_id}")
+                    result = ser.hydrate(
+                        data,
+                        what=f"the result of step {event.correlation_id}",
+                        key=self.run_key,
+                    )
                     if not sus.future.cancelled():
                         sus.future.set_result(result)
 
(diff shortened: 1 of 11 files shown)


src/vercel/_internal/workflow/runtime.py
elprans, line 461: Is `recorded_input` ciphertext? If so, this would always be true?

elprans, line 500: Do we want to hydrate here instead of raising e? I think it'll be ciphertext here?

fantix: Good catch! Yes I was aware of this potential breakage but it's not yet an issue.

The encryption salt changes for each replay, so this must not be ciphertext. This PR is not yet encrypting step inputs, so we are safe for now; when the encryption lands, this check should decrypt first before byte-comparing.

fantix: Yes, it will be! Given that we haven't implemented vercel/workflow#1851 and encryption writing, this is not failing in this PR yet.

elprans (approved):

PR #282, opened 2026-08-11
workflow: carry run creation data on the queue

fantix:
First of three. Groundwork only — no behaviour changes yet.

`start()` writes `run_created` and pushes the queue message with no ordering between them. The TS client issues the two [concurrently on purpose](https://github.com/vercel/workflow/blob/main/packages/core/src/runtime/start.ts) (`Promise.allSettled`, "If events.create fails with 429/5xx, the run was still accepted via the queue"), so a consumer can be handed a run id whose row does not exist yet. The data needed to recover from that rides the message as `runInput` — and we were dropping it: pydantic ignores unknown fields, so it never survived `model_validate`.

## What's here

- **`RunInput` on `WorkflowInvokePayload`** — keeps what a first delivery brings. Re-enqueues omit it, as do producers below spec version 3.
- **`RunStartedEventData`** — `run_started` can now carry that data back as `eventData`, which is how a world creates the run from the message instead of from the missing row. Every field optional, mirroring `RunStartedEventSchema` upstream.
- **`World.get_environment()`** — reports the environment a world's writes are attributed to. Nothing consults it yet.

Attributes and `encryptionPublicKey` are pass-through: this SDK reads neither, but a run recreated without them would lose both permanently, so they are carried verbatim rather than dropped.

`get_environment()` on the Vercel world reads back the `x-vercel-environment` header we already send, falling back to `VERCEL_TARGET_ENV || VERCEL_ENV`. Reading the header back is deliberate — it is what keeps the reported value and the sent one from drifting, the same reason upstream shares one helper between `getHeaders` and `getEnvironment`. `VERCEL_TARGET_ENV` is preferred over `VERCEL_ENV` because the latter reports `preview` inside a Vercel custom environment while the OIDC claim carries the custom environment's slug.

## Also

One pre-existing over-length line in `runtime.py` that fails `ruff` on `main`, wrapped so CI can go green.

## Test plan

16 new tests pinning the wire shape both directions — the producer is the TypeScript SDK, so a dropped field is silent. `uv run poe qa` green.

### DIFF
--- a/src/vercel/_internal/workflow/worlds/vercel.py
+++ b/src/vercel/_internal/workflow/worlds/vercel.py
@@ -419,6 +419,13 @@ class VercelWorld(w.World):
             raise ValueError("VERCEL_DEPLOYMENT_ID environment variable is not set.")
         return deployment_id
 
+    def get_environment(self) -> str | None:
+        """utils.ts, resolveClientEnvironment."""
+        header = self._headers.get("x-vercel-environment")
+        if header:
+            return header
+        return os.getenv("VERCEL_TARGET_ENV") or os.getenv("VERCEL_ENV") or None
+
     async def run_key(self, run_id: str, *, deployment_id: str | None = None) -> bytes | None:
         """Resolve the key that opens *run_id*'s ``encr`` payloads.
 
--- a/changes/vercel/workflow-resilient-start.bugfix.md
+++ b/changes/vercel/workflow-resilient-start.bugfix.md
@@ -1,0 +1,1 @@
+Carry run creation data through the workflow queue for resilient start.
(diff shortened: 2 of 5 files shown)


src/vercel/_internal/workflow/worlds/vercel.py
```diff
@@ -419,6 +419,13 @@ async def get_deployment_id(self) -> str:
             raise ValueError("VERCEL_DEPLOYMENT_ID environment variable is not set.")
         return deployment_id
 
+    def get_environment(self) -> str | None:
```
elprans, line 422: Is this public API?

elprans (approved):

src/vercel/_internal/workflow/worlds/vercel.py
fantix: Negative, such world APIs are currently internal and only consumed by the workflow public APIs.

PR #294, opened 2026-08-13
workflow: let an app serve the workflow endpoint

fantix:
This adds a semi-internal API that allows any Web framework to serve the workflow endpoint for JS-tools integration in local development and e2e test.

```python
from collections.abc import AsyncIterator, Mapping

import fastapi
from vercel.workflow import ENDPOINT_PATH, Workflows

wf = Workflows()
app = fastapi.FastAPI()


class Request:
    def __init__(self, request: fastapi.Request) -> None:
        self._request = request

    @property
    def method(self) -> str:
        return self._request.method

    @property
    def url(self) -> str:
        return str(self._request.url)

    @property
    def headers(self) -> Mapping[str, str]:
        return self._request.headers

    async def aiter_bytes(self, chunk_size: int | None = None) -> AsyncIterator[bytes]:
        async for chunk in self._request.stream():
            yield chunk


@app.api_route(ENDPOINT_PATH, methods=["POST", "GET", "HEAD", "OPTIONS"])
async def workflow_endpoint(request: fastapi.Request) -> fastapi.Response:
    response = await wf.http_handler(Request(request))
    return fastapi.Response(
        content=response.body,
        status_code=response.status,
        headers=response.headers,
    )
```

This will become a implementation detail of the future `workflow.with_workflow(app, wf)` for each supported Web framework. Before that, we expose such intermediate API so that e2e test could work without hacking to use the `vercel.workflow._internal` symbols. This is not advertised in any public docs.


changes/vercel-workflow/workflow-http-endpoint.feature.md
```diff
@@ -0,0 +1 @@
+Serve workflows from your own web framework: `Workflows.http_handler`, `ENDPOINT_PATH`, `HTTPRequest` and `HTTPResponse` are now public.
```
msullivan: Do we want to advertise this?

Or maybe flag it as unstable, if we think it's semi-internal?

msullivan (approved):

changes/vercel-workflow/workflow-http-endpoint.feature.md
fantix: Yeah let me add some notes here - in the long run, I think we want to advertise `vercel.workflow.with_workflow(my_fastapi_app, my_workflow_registry)` instead, which is a framework-aware wrapper of these HTTP primitives.

PR #296, opened 2026-08-13
workflow: support exposing manifest.json

fantix:
This PR generates `manifest.json` of the workflow/step registry in the same format as the TS Workflow SDK does, and expose it in 2 ways:

1. Through HTTP for e2e test use only; this is a semi-internal API and gated behind `WORKFLOW_PUBLIC_MANIFEST=1`
2. `python -m vercel.workflow manifest` writes the JSON to a file `$WORKFLOW_LOCAL_DATA_DIR/manifest.json`, which is read by the JS CLI tools to show what workflows/steps are in this project and their shapes.

A design decision here to review is: the TS tools must detect `pyproject.toml` and see workflow definitions there, and call `python -m vercel.workflow manifest` every time before reading the manifest file. The rational is, a user may just modify the Python files and invoke TS tools to see updates, and there can be no user-triggered Python code execution time in between. An alternative solution would be to integrate with `vc dev` but workflow does not necessarily depend on the Vercel CLI, so this path has a flaw.


msullivan:
> A design decision here to review is: the TS tools must detect pyproject.toml and see workflow definitions there, and call python -m vercel.workflow manifest every time before reading the manifest file.

How does it get the manifest for TS?

src/vercel-workflow/vercel/workflow/_internal/cli.py
```diff
@@ -0,0 +1,171 @@
...
+
+def _declared_entrypoints(project: pathlib.Path) -> list[str]:
+    path = project / PYPROJECT
+    try:
+        with path.open("rb") as f:
+            config = tomllib.load(f)
+    except FileNotFoundError:
+        raise CommandError(
+            f"no {PYPROJECT} in {project}, so there is nothing to read the app from. "
+            f"Name it instead: {PYTHON} -m vercel.workflow manifest module:app"
```
msullivan: "Specify it explicitly", maybe.

It doesn't really matter but Claude is weird about the word "name". (This one is less weird than most of its weird ones though.)

msullivan (approved):
I think the real question about the design decision here is how the TS workflow team feels about it. The tools checking for Python and triggering a manifest build seems mildly distasteful but not terribly so, but we'll have to see if the tool owners agree.

Either way I think this PR is mergeable now? There's probably nothing to do here differently even if they reject pyproject.toml checking.

Alternatives in that case, like you said: we could integrate with `vc dev`. If someone isn't using `vc dev`, then it requries a manual build step.

fantix:
> > A design decision here to review is: the TS tools must detect pyproject.toml and see workflow definitions there, and call python -m vercel.workflow manifest every time before reading the manifest file.
> 
> How does it get the manifest for TS?

The Workflow JS SDK joins the build process of user application and the manifest is simply a built output. The CLI tool just need to pick it up from the filesystem and read directly.

> I think the real question about the design decision here is how the TS workflow team feels about it. The tools checking for Python and triggering a manifest build seems mildly distasteful but not terribly so, but we'll have to see if the tool owners agree.
>
> Either way I think this PR is mergeable now? There's probably nothing to do here differently even if they reject pyproject.toml checking.

Right, let's make the Python end work as-is tentatively and I'll raise a discussion with probably a proposal PR for the CLI.

> Alternatives in that case, like you said: we could integrate with `vc dev`. If someone isn't using `vc dev`, then it requries a manual build step.

Yeah, that sounds like a plan! As well as `fastapi dev` and we can inject something in the `with_workflow()` integration.

PR #299, opened 2026-08-13
Separate workflow into `vercel-workflow`

fantix:
Moves `vercel.workflow` and its internals out of the `vercel` distribution into a new `vercel-workflow` package, published at `0.9.0`.

* `vercel` depends on `vercel-workflow`, so everything still works as-is.
* `vercel-workflow` is excluded from the bundled releases.
* `cryptography>=43.0.0` is now a required dependency of `vercel-workflow`, thus needed by `vercel` too.
* `cryptography>=43.0.0` is the first version that passes our test suite. Ubuntu 24.04LTS comes with `41.0` as a reference. No upper bound so the user could receive latest security updates.


scripts/bundle_release.py
```diff
@@ -83,6 +83,10 @@
     # the `vercel-oidc[verify]` extra rather than this package's own table, and
     # extras are already passed through unvendored.
     "vercel-connect": {"pydantic"},
+    # Pydantic again, and `cbor2` for the same reason: its fast path is the
+    # `_cbor2` C extension, which vendoring drops along with every other `.so`,
+    # leaving a bundle that silently falls back to the pure-Python codec.
+    "vercel-workflow": {"pydantic", "cbor2"},
```
elprans: No need to produce the bundle for this for now, we're not planning on vendoring it.

src/vercel-workflow/pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+dynamic = ["version", "dependencies"]
+description = "Python SDK for Vercel Workflows"
+readme = "README.md"
+requires-python = ">=3.10"
+license = "MIT"
+license-files = ["LICENSE", "LICENSE.*"]
+
+[project.optional-dependencies]
+encryption = [
+    "cryptography>=48.0.1",
```
elprans: Let's make this non-optional, but find a reasonable version floor so we don't induce peer conflicts.

elprans (approved):

src/vercel-workflow/vercel/workflow/version.py
```diff
@@ -0,0 +1,3 @@
+"""Package version metadata."""
+
+__version__ = "1.0.0b1"
```
elprans: Let's not do prereleases, 0.9.0 will do.

scripts/bundle_release.py
fantix: Roger; added `UNBUNDLED_PACKAGES` to opt-out.

PR #300, opened 2026-08-14
workflow: deliver a hook resume carried on the queue

fantix:
This is similar to the [resilient start](https://github.com/vercel/vercel-py/pull/284), and mirrors vercel/workflow#1834 with a few [following fixes](https://github.com/vercel/workflow/pull/3345).

In short, a vqs message to resume a hook can now also carry the actual payload, in addition to the `hook_received` workflow server event. And these 2 things may arrive without a particular order for performance, so the SDK must be able to handle all cases properly.

This would unblock the hook-based e2e test fixtures. This PR does not introduce any public API.

Currently, the `HOOK_RESUME_INPUT_VERSION` cannot protect the Python app from e2e test client issuing parallel hook resume calls: TS SDK honors peer's `hookResumeInputVersion` only for cross-deployment calls, but most e2e tests are same-deployment calls. With that said, we should probably teach SDKs about cross-SDK awareness, after we really open up cross-language workflow calls on top of cross-deployment calls (e2e test is currently an exception for just ourselves).

---

## The bug

Resuming a hook takes two writes: the `hook_received` event, and the queue message
that wakes the run. `resumeHook()` can do both **at once** and put the payload on
the message (vercel/workflow#3230 made that the default), so a delivery can arrive
before the event exists.

We never read that payload. The replay found nothing new, suspended, and answered
200:

```
run_created, run_started, hook_created, wait_created     ← log at replay time
hook_received                                            ← producer's write, landed after
```

Nothing was left to wake the run: exactly one message is published per resume, and
this delivery consumed it. **The resume was lost, not delayed.**

In the TypeScript e2e suite `hookWithSleepWorkflow` stalled on its first payload
every time; `hookWithSleepFinalStepWorkflow` escaped only because its driver sends
a second payload, whose delivery replays the whole log and repairs the miss.
Setting `WORKFLOW_DISABLE_LAZY_HOOK_RESUME=1` on the driver made both pass, which
is what pinned the cause.

## The fix

| Layer | Change |
|---|---|
| `world.py` | `hookInput` on the invoke payload; `token` on `hook_received`; `resumeId` on the stored event; `HOOK_RESUME_INPUT_VERSION` |
| `runtime.py` | write the event before replay when the log lacks it; advertise the version from `start()` and from the health-check answer |
| `worlds/vercel.py` | forward the resume identity on the request and let the backend enforce it |
| `worlds/local.py` | no backend to defer to, so a `(runId, resumeId)` claim file, byte-shaped like `world-local`'s |
| `ulid.py` | `decode_time()`, so the event is dated when the resume happened rather than when we got to it |

A delivery that has to write the event carries the `hookInput` it arrived with on
the event itself, as a private attribute, so `events_create` keeps its signature
and only the two worlds that act on it need to look. Nothing about it reaches the
wire; each world copies out the fields it needs.

Two writers of one resume must collapse into **one** event, or the workflow body
gets the payload twice — worse than the stall. That is what the identity is for.

Of the three conditions the claim can fail, two — a resume id already recorded for
a different hook, or for a different payload — read the same on every redelivery,
so they raise `HookResumeConflictError` and the consumer drops the message with a
log saying what is wrong. Only "claimed, event not readable yet" is retried.
Upstream retries all three; it can afford to, because it stops at 48 deliveries.

## Two divergences from `world-local`, of different weight

**1. A claim whose event has not landed yet — observed.** Upstream *takes over*
the position, reasoning that only a crashed writer leaves one behind. That holds
while both writers share the process its per-hook lock serializes
(`withInProcessLock`, `events-storage.ts:974`); across processes it does not. The
other writer numbers events by position, so finding ours where it meant to publish,
it bumps to the next free slot instead of failing. **Two `hook_received` events for
one resume, in three of eight suite runs.** We raise instead, costing one
redelivery. What that gives up is a writer that really did crash between its two
writes.

**2. `_is_resume_event` matches on the resume id only — inferred, not observed.**
Upstream also accepts an event with *no* resume id at the claimed position, reading
it as one written before the field existed. But the claim file (vercel/workflow#3230)
post-dates that field (vercel/workflow#1834), so nothing a claim can point at
predates it — the branch guards data that cannot exist, while admitting an
unrelated payload for the same hook that happens to sit at the claimed position.
Removing it costs nothing and turns a silent wrong-event match into a loud retry.
**No such match has been observed**, here or upstream, and it has not been reported
upstream; the test that covers it builds the claim by hand.

## Verification

- **`29 passed | 108 skipped`** against the TypeScript e2e driver, up from 27, with
  both `hookWithSleep*` fixtures off `unsupported` and zero duplicated resumes
  across the logs of six runs.
- A new integration test drives a real `LocalWorld` with a producer that publishes
  before it writes. Reverting the handler hunk fails all three of its tests with
  the original signature (`log is ['run_created', 'run_started', 'hook_created']`).
- Unit coverage split by layer, matching the resilient-start tests: models,
  handler, local world.
- `py3.10` runs the whole package under plain `pytest` in one process, which is
  what CI does and what a 14-worker local `ggt` run hides: the integration
  fixture's world is opened and closed by the test's own task, since anyio will
  not let another task exit that cancel scope.

## Not in scope

- **No `MAX_QUEUE_DELIVERIES`.** Upstream caps every delivery chain at 48 and then
  fails the run (`runtime.ts:727`); we have no such cap and no `max_attempts` on
  the subscription, so *any* exception escaping `workflow_handler` redelivers
  forever. Pre-existing and wider than this PR — the resume path works around its
  own share of it (see `HookResumeConflictError` above), nothing else does.
- `hookInput.deploymentId` is parsed and unused; upstream re-routes a misrouted
  delivery with it, which this SDK does for no message type.
- `resume_hook()` still writes `hook_received` without `token`. Harmless while it
  is the only writer of its event, but it is the same lossy-parsing family as the
  other hook gaps.
- Found on the way: a log containing a created-event the body never issues makes
  `resume()` spin forever instead of failing. Pre-existing, filed separately.


src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
```diff
@@ -408,6 +418,101 @@ def _new_id(self, prefix: str) -> str:
...
+        key = hashlib.sha256(f"{run_id}\x00{resume_id}".encode()).hexdigest()
+        return self.data_dir / "hooks" / "resumes" / f"{key}.json"
+
+    def _read_event(self, run_id: str, event_id: str) -> w.Event | None:
+        return read_json(self.data_dir / "events" / f"{run_id}-{event_id}.json", w.EventAdaptor)
+
+    @staticmethod
+    def _is_resume_event(event: w.Event, claim: HookResumeClaim) -> bool:
+        # The event has to say it is this resume. `world-local` also accepts one
+        # with no resume id at all, on the grounds that it predates the field --
```
msullivan: Does this make sense?

src/vercel-workflow/vercel/workflow/_internal/world.py
```diff
@@ -108,6 +108,26 @@ def get_physical_topic(queue_name: str) -> SanitizedName:
...
+# It says so by number, and a run records the number in its `executionContext`
+# when it is created. `resumeHook()` reads it off the run and only takes the
+# fast path when it is high enough; a run without one gets the safe path.
+#
+# This is not a spec version. Those describe how a run's rows are encoded, which
+# a reader can see in the rows. This describes what *code* will consume the
+# run's queue messages, which nothing in the data reveals.
+#
+# Mirrors `HOOK_RESUME_INPUT_VERSION` in `@workflow/world`'s `hooks.ts`.
+HOOK_RESUME_INPUT_VERSION = 1
```
msullivan, line 131: If the hookResume code consumes this, and that decides whether to do this, then really all of this is more an optimization than a bugfix, right, because the parallel path shouldn't have been firing? So then why were tests failing?

What am I missing?

msullivan: Oh, because there is also the previously existing case where creating the hookResume event *failed*, and workflow decided to just ignore the failure and let the queue handle it? And that's what one of the tests was testing I guess?

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -735,6 +735,91 @@ def resume(self) -> None:
...
+    Safe to call when the other writer of this resume also wrote the event, or is
+    writing it right now: the world is told which resume this is
+    (:class:`~world.HookResume`) and collapses the two into one event.
+
+    Note what it does *not* catch. Every other write in this module swallows
+    ``EntityConflictError``; here it is allowed out. A conflict means the other
+    writer has claimed this resume but its event is not readable yet, which is
+    temporary -- so the message goes unacked and a later delivery finds the
+    event. Swallowing it would ack a message that may hold the only copy of the
+    payload.
```
msullivan: OK, and so then the workflow invocation will fail and get retried?

src/vercel-workflow/vercel/workflow/_internal/world.py
```diff
@@ -1037,20 +1084,27 @@ async def events_create(self, run_id: None, data: RunCreatedEvent) -> EventResul
...
+                resume.
         Returns:
             The created event and affected entity
         """
         ...
 
     @abc.abstractmethod
-    async def events_create(self, run_id: str | None, data: Event) -> EventResult: ...
+    async def events_create(
+        self, run_id: str | None, data: Event, *, resume: HookResume | None = None
```
msullivan: Adding a `resume` argument seems suuuuper ad-hoc. I think it would be better to add HookResume as a non-serialized optional field to HookReceivedEvent

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
```diff
@@ -408,6 +418,101 @@ def _new_id(self, prefix: str) -> str:
...
+        def converge(claim: HookResumeClaim) -> w.Event:
+            # Same resume id, different hook. The two writers of one resume always
+            # agree on which hook it is, so this is a caller bug (or an id
+            # collision) rather than the same resume arriving twice. Reusing that
+            # hook's event here would file this payload under the wrong hook.
+            if claim.hook_id != data.correlation_id:
+                raise w.EntityConflictError(
+                    f'hook_received resumeId "{resume.resume_id}" already recorded '
+                    "for a different hook"
+                )
```
msullivan, line 479: Are these legitimate errors that might actually happen? And are probably fine I guess because EntityConflictError is usually swallowed? Or do they indicate something really weird is wrong? (And thus should maybe be asserts?)

msullivan (approved):

src/vercel-workflow/vercel/workflow/_internal/world.py
fantix: > If the hookResume code consumes this, and that decides whether to do this, then really all of this is more an optimization than a bugfix, right, because the parallel path shouldn't have been firing? So then why were tests failing?
> 
> What am I missing?

Ah, `hookResume()` consumes the version yes, but not directly the one we defined here, but rather the `hookResumeInputVersion` pinned during the `start()` call.

TS `start()` only honors `hookResumeInputVersion` in a cross-deployment call through health check results; e2e test (both local and Vercel deployment) uses the same `deployment_id` in the client as the peer, thus skips probing the peer, and simply assumes the same `hookResumeInputVersion` as the TS SDK itself, which is version 1 by default (a.k.a. parallel hook resume is supported), and requires the fix in this PR.

Also highlighted in PR description now.

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
fantix: The first two `EntityConflictError`s here are caller bugs; they are very wrong and shouldn't happen. But we as the consumer should not panic on client bugs - swallowing the error, logging and ack-ing the queue event is what we *need* to do here.

Raising `EntityConflictError` here means redelivery, and it's 1:1 TS behavior, which now I think is wrong in the first place. I will replace them with a different error type and handle them without redelivery.

The third `EntityConflictError` is a legitimate error that may happen (we failed in a claim race, but couldn't see the winner's event either). We let that propagate and end up with redeliver on purpose, expecting to see the winner's event in the next replay.

src/vercel-workflow/vercel/workflow/_internal/world.py
fantix: Fixed!

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Yes exactly; docstring reworded.

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
fantix: Looks like it was only trying to explain why it diverged from the JS implementation. Reworded.

PR #302, opened 2026-08-15
workflow: retry a step at the time it asks for

fantix:
This PR adds a new public API - a new error type `vercel.workflow.RetryableError`. 

```python
from vercel.workflow import FatalError, RetryableError


@app.step(max_retries=5)
async def charge_customer(customer_id: str) -> None:
    if not customer_id:
        # Retrying cannot help, so fail the step on this attempt.
        raise FatalError("no customer")
    response = await http.post(...)
    if response.status_code == 429:
        # Retry, but not before the API says we may.
        raise RetryableError("rate limited", retry_after="10s")
```

`RetryableError("rate limited", retry_after="10s")` sets how long to wait before a step's next attempt, mirroring `RetryableError` in TS `@workflow/errors`. `retry_after` accepts what `sleep()` accepts and defaults to one second. It changes when the next attempt runs, not how many attempts there are.

The deadline is recorded twice in `_execute_step()` in `runtime.py`. The `QueueContinuation.delay_seconds` delays this delivery; the `StepRetryingEventData.retryAfter` persisted on the step is used by the world (like workflow server) to prevent staring the step too early (returning `TooEarlyError` if so).


src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
```diff
@@ -1064,21 +1074,31 @@ def _events_create_impl(self, run_id: str | None, data: w.Event) -> w.EventResul
                 )
                 write_json(step_path, step, overwrite=True)
 
+        elif data.event_type == "step_retrying" and hasattr(data, "event_data"):
```
msullivan, line 1077: `hasattr` is a smell; why is it needed here?

src/vercel-workflow/vercel/workflow/_internal/duration.py
```diff
@@ -0,0 +1,55 @@
...
+    "d": 24 * 60 * 60 * 1_000,
+    "day": 24 * 60 * 60 * 1_000,
+    "days": 24 * 60 * 60 * 1_000,
+    "w": 7 * 24 * 60 * 60 * 1_000,
+    "week": 7 * 24 * 60 * 60 * 1_000,
+    "weeks": 7 * 24 * 60 * 60 * 1_000,
+}
+
+
+def parse_duration_to_date(param: int | float | datetime | str) -> datetime:
```
msullivan, line 34: I guess none of the comment I'm about to make is really a review for *this* PR, since this is just a refactor, but.

First something we should definitely change:
* It should support taking `timedelta` also. (And as a real nitpick, `duration_units` ought to be defined as timedeltas)

Then some other things that we could (and I think should, probably?) change but would be breaking:
* int/float arguments ought to be seconds, not ms. time.sleep and asyncio.sleep take seconds, and we ought to match.
* I'm not sure good idea to support `datetime`. I think that waiting for a duration and waiting until a time are different operations and I am a little suspicious of distinguishing them by type overloading (esp because int and float *can* represent a fixed point in time also! `time.time()` returns a float...). If we dropped it though, we would have to add `sleep_until` that takes datetime...

An old-man-yells-at-cloud gripe rather than a real change suggestion:
* I think it's a little cheesy to support the strings as a builtin thing, though I understand the appeal. Having a helper function would be cleaner, I guess.

I assume that all three of those last complaints are just us matching the TS SDK, though. For the int/float as ms one, though, I think that matching JS is a mistake, and we should try to do the more idiomatic python thing. (setTimeout/setInterval in JS take ms, but sleep/call_later in python take s.)

src/vercel-workflow/vercel/workflow/_internal/errors.py
```diff
@@ -1,5 +1,43 @@
...
+    string, a number of milliseconds, or an absolute timezone-aware
+    ``datetime``. It defaults to one second from now, as ``RetryableError``
+    in ``@workflow/errors`` does.
+
+    It changes only *when* the next attempt runs, not how many attempts there
+    are: a step that has used up its ``max_retries`` fails whether or not it
+    raised this.
+    """
+
+    retry_after: datetime
```
msullivan: This is `retry_at`, really

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1409,26 +1416,33 @@ async def _execute_step(
...
+
             # Set step to pending for retry
             error_stack = traceback.format_exc()
             await world.events_create(
                 req.run_id,
-                w.StepRetryingEventData(error=error_text, stack=error_stack).into_event(
-                    req.step_id
-                ),
+                w.StepRetryingEventData(
+                    error=error_text, stack=error_stack, retryAfter=retry_after
```
msullivan:
```suggestion
                    error=error_text, stack=error_stack, retry_after=retry_after
```

msullivan: We should use the pythonic name, not the alias, I think.
(I actually didn't realize you even *could* use the alias -- I thought it was just for serialization!)

msullivan (approved):
Looks great.

Mostly I just ranted about an unrelated issue this touches on, which is the time arguments taken by `sleep` and `retry_after`. We should sync up on that out of band probably. Doesn't really affect whether to merge this now

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
fantix: This was a long-time copy-over from the TS code that is not needed in Python. Let me remove them altogether in a follow-up PR.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Yeah, this was a very early decision point about using `alias` on `pydantic.Field()` for all those who have camelCase on the wire but snake_case in Pyhton. `alias` actually applies to both validation and serialization (with our global `serialize_by_alias=True` model_config), which is needed for the wire, but unfortunately Pydantic object construction is also validation. There is the `validate_by_name=True` config but it also opens up snake_case for wire fields so I didn't turn that on in the first place and kept the camelCases in Python object constructions.

It's possible to globally set `validate_by_name=True` + `validate_by_alias=False`, but only call `model_validate()` with `by_name=False` + `by_alias=True`. That would eliminate all camelCase Python code. Let me start a new PR for that.

src/vercel-workflow/vercel/workflow/_internal/duration.py
fantix: Yes, I agree! It was a JS-matching thing and we should change that. @scotttrinh mentioned [something similar](https://github.com/vercel/vercel-py/pull/109#discussion_r3119940904). Let me also start a new PR.

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
fantix: Fixed in #321

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Fixed in #322

PR #303, opened 2026-08-15
workflow: set a run's attributes from a workflow or a step

fantix:
This PR adds 3 new public APIs:

1. `vercel.workflow.set_attributes()` writes plaintext user key/value attributes onto the run
2. `vercel.workflow.remove_attributes()` drops attributes by keys
3. `vercel.workflow.Run.attributes()` fetches the current attributes of the run

1 and 2 can only be called within a workflow run or a step, while 3 is only available on the client-side. In other words, workflows and steps themselves can only write attributes, for the caller to read.

For example, a workflow (or a step):

```python
from vercel.workflow import remove_attributes, set_attributes

@app.workflow
async def renew_subscription(customer_id: str) -> None:
    await set_attributes(customer=customer_id, phase="charging")
    await charge_customer(customer_id)
    await set_attributes(phase="done")
    await remove_attributes("customer")
```

And the caller:

```python
from vercel.workflow import start

run = await start(renew_subscription, "cus_123")
await run.return_value()

assert await run.attributes() == {"phase": "done"}
```


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1212,9 +1382,10 @@ async def dispose_hook(h=hook):
                 tg.start_soon(dispose_hook)
                 events_created = True
 
-    if not context.suspensions and events_created:
-        # We captured a _SuspendException but there is no suspension - this is likely caused
-        # by a disposed hook that cleared its suspensions. Just retry if event log changed.
+    if attributes_written or (not context.suspensions and events_created):
```
fantix, line 1478: `attributes_written` is future-insufficient; `AttrSetEvent` is not the only event that - once written - is immediately eligible for another in-process replay. `HookConflictEvent`, `StepFailedEvent` are examples (each leading to a separate bug) of this branch. I shall promote `attributes_written` into a `replay_ready` flag or even a `FlushResult` class in a following PR.

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
```diff
@@ -1013,6 +1018,40 @@ def _events_create_impl(self, run_id: str | None, data: w.Event) -> w.EventResul
...
+                claim_path = (
+                    self.data_dir
+                    / ".locks"
+                    / "attributes"
+                    / f"{effective_run_id}-{data.correlation_id}.created"
+                )
+                if not write_exclusive(claim_path, ""):
+                    raise w.EntityConflictError(
+                        f'Attribute event "{data.correlation_id}" already exists'
+                    )
```
msullivan: I'm not sure I follow this

src/vercel-workflow/vercel/workflow/_internal/attributes.py
```diff
@@ -0,0 +1,92 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+
+RESERVED_ATTRIBUTE_KEY_PREFIX = "$"
+"""Reserved for framework and library code; user code has to opt in."""
+
+ATTRIBUTE_KEY_MAX_LENGTH = 256
+"""Max length of an attribute key, in UTF-16 code units."""
```
msullivan, line 12: Hate JS.

msullivan (approved):

src/vercel-workflow/vercel/workflow/_internal/worlds/local.py
fantix: Ahh I missed to update this agent rant. It's trying to explain why `validate_attribute_changes` must happen before claiming the correlation ID - to avoid a claimed-but-failed-validation case. Step writes don't need claims because they don't get concurrently replayed.

Reworded.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Fixed in #327

PR #307, opened 2026-08-19
sandbox: Skip httpx client timeout for unbounded operations

scotttrinh:
- Add support for "no timeout" value to pass to the transport's client on a per-call basis. `None` was already taken to mean "use the default", so instead of a special value like `timedelta(seconds=0)`, I went with a sentinel object for expressing no timeout.
- Use this new no-timeout request for undefined/unbounded latency operations on Sandbox (e.g. command streaming). The further fix here is to wrap these operations in a durable, long-running, fault tolerant wrapper, but for now this will hold open the connection as long as it can.

### DIFF
--- a/src/vercel-internal-core/vercel/_internal/core/http/transport.py
+++ b/src/vercel-internal-core/vercel/_internal/core/http/transport.py
@@ -64,6 +64,16 @@ class RawBody:
 RequestBody = JSONBody | BytesBody | RawBody | None
 
 
+class _NoTimeout:
+    """Select no HTTPX timeout instead of inheriting the client default."""
+
+    __slots__ = ()
+
+
+NO_TIMEOUT = _NoTimeout()
+RequestTimeout: TypeAlias = timedelta | _NoTimeout | None
+
+
 class ReadResponsePolicy(StrEnum):
     ALWAYS = "always"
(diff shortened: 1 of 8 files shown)


src/vercel-internal-core/vercel/_internal/core/http/transport.py
```diff
@@ -64,6 +64,16 @@ class RawBody:
 RequestBody = JSONBody | BytesBody | RawBody | None
 
 
+class _NoTimeout:
+    """Select no HTTPX timeout instead of inheriting the client default."""
+
+    __slots__ = ()
+
+
+NO_TIMEOUT = _NoTimeout()
```
elprans:
```suggestion
NO_TIMEOUT: Final = _NoTimeout()
```

src/vercel-internal-core/vercel/_internal/core/http/transport.py
```diff
@@ -64,6 +64,16 @@ class RawBody:
 RequestBody = JSONBody | BytesBody | RawBody | None
 
 
+class _NoTimeout:
```
elprans, line 68:
```suggestion
@final
class _NoTimeout:
```

elprans (approved):

PR #310, opened 2026-08-20
workflow: support using a shared sandbox for all workflow invocations

msullivan:
Importing everything on every workflow invocation is quite slow,
especially when using pydantic or similar.

I am planning to add automatic support for using pydantic to
serialize/validate arguments, and regenerating the pydantic validators
on every invocation will make it even more expensive. Using a shared
sandbox eliminates that problem.

Using a shared sandbox means that we still get all of the benefits of
preventing nondeterminstic calls, but that modifications to global
variables will stay visible.
But I think a good workflow shouldn't really be using globals
*anyway*, so...

I think I'm planning to basically immediately make a PR that will
switch the default to `share_sandboxes=True`, but for cleanliness I'm
separating it.

To do this, I:
 * Add a Sandbox class entered with enter()
 * Add a policy flag for it
 * Make `Workflows` responsible for producing a sandbox
 * Hold a sandbox-scoped lock while *importing* the module, to avoid
   weird cyclic init races.
 * A bunch of tweaks to serde's registries

Questions:
 * Are we nervous about sharing a sandbox for genuinely concurrent
   runs? If so, we can maintain a pool of unused sandboxes and create
   a new one when necessary.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
@@ -34,8 +34,8 @@ _in_sandbox: contextvars.ContextVar[bool] = contextvars.ContextVar("_in_sandbox"
 # and writes this dict directly, so non-workflow code is unaffected.
 _real_sys_modules: dict[str, types.ModuleType] = sys.modules
 
-# Per-execution module table.  A workflow run sets this to its own private dict
-# (see workflow_sandbox) so concurrent runs — whether on different asyncio
+# Per-sandbox module table.  Each sandbox sets this to its own dict (see
+# Sandbox.enter) so runs in different sandboxes — whether on different asyncio
 # tasks or different threads — never share or clobber each other's modules.
 # ``None`` means "use the real table".
 _sandbox_sys_modules: contextvars.ContextVar[dict[str, types.ModuleType] | None] = (
@@ -692,11 +692,8 @@ class _DispatchingSysModules(MutableMapping[str, types.ModuleType]):
     """Installed once as ``sys.modules``.
 
     Every read/write dispatches to the current context's module table: a
-    workflow run's private dict while a sandbox is active in this context
-    (asyncio task or thread), or the real process table otherwise.  The real
-    dict object is never cleared, so concurrent workflows on different tasks or
-    threads can neither corrupt each other nor the host — which the previous
-    ``sys.modules.clear()`` approach could not guarantee.
+    sandbox's private dict while a sandbox is active in this context
+    (asyncio task or thread), or the real process table otherwise.
     """
 
     __slots__ = ()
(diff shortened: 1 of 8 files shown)


fantix:
> Are we nervous about sharing a sandbox for genuinely concurrent runs? If so, we can maintain a pool of unused sandboxes and create a new one when necessary.

Each workflow run in the shared sandbox is sequential and serialized, right? Like workflow A starts an event loop, run to suspension, shutdown that loop, THEN workflow B can only start a new event loop after that, and do B things? I think I'm okay with shared sandboxes for this case.

src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
```diff
@@ -926,61 +925,98 @@ def clear_pydantic_generics_cache(context: SandboxCleanupContext) -> None:
...
+    A handler that raises is logged and skipped, and never masks the
+    workflow's own exception.  Handlers must be thread-safe: other runs
+    may be executing concurrently.
+    These are *not* run when share_sandboxes=True.
     """
 
-    cleanups: tuple[CleanupHandler, ...] = ()
     passthrough_modules: frozenset[str] = frozenset()
+    share_sandboxes: bool = False
+    cleanups: tuple[CleanupHandler, ...] = ()
```
fantix, line 952: Maybe add `@dataclass(kw_only=True)` for this re-ordering, in case anyone passed positional arguments before this change?

fantix (approved):

msullivan:
> > Are we nervous about sharing a sandbox for genuinely concurrent runs? If so, we can maintain a pool of unused sandboxes and create a new one when necessary.
> 
> Each workflow run in the shared sandbox is sequential and serialized, right? Like workflow A starts an event loop, run to suspension, shutdown that loop, THEN workflow B can only start a new event loop after that, and do B things? I think I'm okay with shared sandboxes for this case.

We will run multiple workflows in separate threads, so they can be genuinely concurrent. A and B in the same sandbox at the same time.

If we are nervous about that, we can make it so that a sandbox is in use by *one* thread at a time, and can then get reused later.

fantix:
> We will run multiple workflows in separate threads, so they can be genuinely concurrent. A and B in the same sandbox at the same time.
> 
> If we are nervous about that, we can make it so that a sandbox is in use by _one_ thread at a time, and can then get reused later.

I feel like it's probably fine with the asyncio `Runner` thread-local thing, as far as each event loop is utilized sequentially for all runs within each thread.

msullivan:
> > We will run multiple workflows in separate threads, so they can be genuinely concurrent. A and B in the same sandbox at the same time.
> > If we are nervous about that, we can make it so that a sandbox is in use by _one_ thread at a time, and can then get reused later.
> 
> I feel like it's probably fine with the asyncio `Runner` thread-local thing, as far as each event loop is utilized sequentially for all runs within each thread.

Yeah, the event loop isn't shared.

PR #312, opened 2026-08-20
workflow: fix calls with positional-or-keyword params and `*args`

msullivan:
When normalizing arguments we try to prioritize binding everything to
a keyword argument, if possible, including positional-or-keyword.

This doesn't work if you have something like:
```
def f(x, **kwargs): ...

f(0, 1)
```
because after roundtripping it will get called like `f(1, x=0)` and
will error because both will bind to `x`.

Fix this by keeping keyword-or-positional arguments as positional if
there is anything bound to `*args`.

Question: should we instead just bias towards binding everything
positionally, like `Signature.bind` will do by default?

### DIFF
--- a/changes/vercel-workflow/variadic-argument-binding.bugfix.md
+++ b/changes/vercel-workflow/variadic-argument-binding.bugfix.md
@@ -1,0 +1,1 @@
+Fixed workflow and step calls with both positional-or-keyword parameters and `*args` failing during replay because their arguments were recorded in an unbindable shape.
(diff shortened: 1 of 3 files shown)


fantix (approved):
Good catch!

> Question: should we instead just bias towards binding everything positionally, like `Signature.bind` will do by default?

I'm not sure - given that this is only a wire-format detail, I think both are fine as far as the behavior is bug-free and stably pinned by tests. Keyword arguments could theoretically survive positional-or-keyword argument re-ordering, but positional arguments would also survive renaming. However, neither is a real problem because code is not supposed to change within a single deployment. It's not a problem for future cross-language calls either - I think we will generate target language binding code that translates whatever wire-format to the target API / function signature, so we will be able to do the proper adaption there.

PR #317, opened 2026-08-21
workflow: use pydantic to validate workflow/step types

msullivan:
The big win here is being able to automatically validate/dump pydantic
models and dataclasses.

A value that does not match raises TypeValidationError, a FatalError:
the bytes a retry would replay are the ones that did not match.

Run is generic now, so return_value() is typed and can validate against
the workflow's own return annotation.

---


src/vercel-workflow/vercel/workflow/_internal/signature_codec.py
```diff
@@ -0,0 +1,202 @@
...
+            for key, value in kwargs.items()
+        }
+
+        return validated_args, validated_kwargs
+
+
+def _invert_bindings(
+    sig: inspect.Signature, args: tuple[Any, ...] | list[Any], kwargs: dict[str, Any]
+) -> dict[int | str, str]:
+    bound = sig.bind(*tuple(range(len(args))), **{key: key for key in kwargs})
```
fantix, line 192: As we are here - maybe capture the `TypeError` of `sig.bind()` and re-raise a fatal error like what's done in `SignatureCodec._validate()`, so that e.g. wrong number of arguments won't get retried. Like `SignatureCodec._bind(*args, **kwargs) -> BoundArguments`?

fantix (approved):
LGTM!

PR #319, opened 2026-08-23
workflow: read spec 7 and skip sealed noops

pranaygp:
Python half of vercel/workflow#3634 (*Add support for 'noop' event type — spec version 7*).

## Why this is urgent for us

That PR moves `SPEC_VERSION_CURRENT` and `SPEC_VERSION_MAX_SUPPORTED` to 7 and switches both `@workflow/world-local` and `@workflow/world-vercel` to `mintedSpecVersion()`, which returns 7 by default. Our reader ceiling was 6, and it is a pydantic `le=` on `BaseEvent.specVersion` — so a spec-7 row does not degrade, it fails validation, and with it the whole page. The `e2e-python` conformance lane runs the TypeScript driver against `workbench/python` over a shared `world-local` data dir, so it is the first thing that breaks: the driver's `run_created` is now labelled 7 and the Python app cannot parse the log it is supposed to replay.

## What spec 7 actually asks of a client

On the server, a spec-7 run's slot positions come from a per-run counter handed out *before* the write that fills them, so concurrent writers never race for a position. The price is that a writer that takes a position and dies leaves a hole, and the backend's read path fills provably-abandoned holes with a server-written `noop` so returned pages stay dense prefixes.

So the client contract is one thing, and it is read-only: **a reader must know a `noop` occupies its slot and means nothing.**

There is no writer half for us. A World only owes the sealing half if it pre-assigns positions ahead of the commit; our local world allocates each id at the commit that fills it, so it has no holes to seal and will never emit one.

## The change

**`world.py`**

- `SPEC_VERSION_SUPPORTS_SEALED_LOG = 7`, and `SPEC_VERSION_MAX_SUPPORTED` follows it. `SPEC_VERSION_CURRENT` stays 2 — we still stamp what we always stamped, and a run created elsewhere keeps its creator's version.
- `NoopEvent` / `NoopEventData`, in the read union and deliberately **not** in `CreateEventRequest`. `eventData` is optional and open (`extra="allow"`, matching the `.passthrough()` on the TS schema): nothing reads it, the shape belongs to whichever backend sealed the slot, and a reader whose only interest is skipping the row has no business rejecting a field it has not heard of.
- `is_sealed_noop_event()`, mirroring `isSealedNoopEvent` — one home for the test.

**`runtime.py`** — `get_all_workflow_run_events()` drops seals, so the log the replay walks is the log without them.

That is the whole skip, and the reason it is one line at the loader rather than a `continue` in `resume()` is the clock. Everything downstream of that function is reconstructing what the run did — matching correlation ids to suspensions, looking for a terminal event, deciding which waits elapsed, and reading `createdAt` for `now()`. Nothing on this side walks positions, so there is no caller that wants the seal. Meanwhile a seal's `createdAt` is the *sealer's* wall clock, which can postdate every real event around it; `now()` dates the run from the last event the replay consumed, so a seal left in the list hands the body a time no event of the run happened at — and then, because the clock only moves forward, every later `now()` too. Dropping it at the boundary makes the equivalence the contract is about (a sealed log replays exactly like the same log whose holes their own writers filled) hold by construction, rather than by three call sites remembering. The cursor is untouched: it is the World's, and still points past every row that was read.

## Tests

`tests/unit/test_workflow_sealed_log_noop.py`, 12 tests in four parts:

- **the row** — parses out of the read union with `eventData`, without it, and with an `eventData` field we have never heard of; the predicate answers for seals only; `NoopEvent` is in `Event` and absent from `CreateEventRequest`; 7 is the ceiling and 8 is still rejected.
- **storage** — a seal round-trips through the local world and numbering continues past it, mirroring `slot-identity.test.ts`'s `stores, lists, and numbers past a noop event`.
- **the loader** — seals at head, middle (consecutive) and tail are dropped with the order of everything else intact; a page of nothing but seals still pages on; the cursor survives.
- **replay** — the real claim, driven end to end through `workflow_handler`: one recorded two-sleep run, twice, differing only in seals at the head, in the middle (including one splitting a `wait_created`/`wait_completed` pair), and at the tail. Same output, same events written back, and the three `now()` readings pinned literally so a change that broke both logs the same way cannot slip past the comparison. Plus the degenerate all-seals-before-the-run shape, where the first `now()` has to find the first *real* event.

Every seal in the file is stamped a minute into the future, so each of those clock assertions would pass by accident if a seal's `createdAt` were plausible. Reverting the one-line filter fails 5 of the 12.

`test_workflow_local_world_format.py`'s two spec-version tests move 6 → 7; the ceiling assertion now derives its match from `SPEC_VERSION_MAX_SUPPORTED` instead of hardcoding the number, so the next bump only has to touch the constant.

Full suite: 1127 tests, 0 failures (`test_the_api_token_is_config_then_env_then_oidc` fails locally on a box with `VERCEL_TOKEN` exported, before and after this change). Lint and typecheck clean repo-wide.

## Follow-up, not in this PR

`workbench/python/pyproject.toml` in `vercel/workflow` resolves `vercel-workflow` from PyPI (0.9.0 today) regardless of its `[tool.uv.sources]` rev, so the `e2e-python` lane only picks this up once it is released and the lock is refreshed there — or once that file git-pins `vercel-workflow` at a commit containing it, which is what the comment block above that source entry describes.


fantix (commented):
LGTM; let me trim off some AI in-code chats and merge this to unbreak the e2e test.

fantix (approved):

PR #320, opened 2026-08-24
workflow: add a get_workflow_metadata function

msullivan:
Callable from both workflows and steps.

Fields mostly match TS.

One annoying mismatch: started_at requires fetching the run in order
to get the information, which we don't current do, and which doesn't
seem worth it to add just for this, at this point?

So started_at is optional and not populated in steps.

Alternatives would be:
 * Don't include started_at at all
 * Don't support calling get_workflow_metadata() from steps


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1331,11 +1426,32 @@ async def _execute_step(
...
+                step_name=step.name,
+                attempt=current_attempt,
+                step_started_at=step_run.started_at,
+            ),
+            workflow_info=WorkflowInfo(
+                run_id=req.run_id,
+                workflow_name=_workflow_name_from_queue(queue_name),
+                # We have no way to get started_at without fetching
+                # the run, which we don't want to have to do.
+                started_at=None,
```
fantix, line 1444: Maybe we can just use `start_result.run.started_at`? My agent read the server code and suggested that; though the TS client SDK is not using this run info from step start result (TS retrieved run for extra info but we don't need to copy that).

fantix (approved):
I think it is okay to have an optional `started_at` for now.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
msullivan: It looks like workflow-server does always populate the `run` field, though the comments in the world docs in TS don't seem to require it.
(Neither the TS local world nor our local world populate run for a step create, though that would be easy to fix)

PR #322, opened 2026-08-24
workflow: refactor to eliminate camel cases in code

fantix:
This is an internal refactor only that replaces all camel case schema constructor parameters with snake cases. The 2nd commit is 100% renamings.

[ref](https://github.com/vercel/vercel-py/pull/302#discussion_r3824935585)

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/world.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/world.py
@@ -451,7 +451,7 @@ class RunCreatedEventData(BaseModel):
     )
 
     def into_event(self) -> "RunCreatedEvent":
-        return RunCreatedEvent(eventData=self)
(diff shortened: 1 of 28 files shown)


src/vercel-workflow/vercel/workflow/_internal/world.py
```diff
@@ -133,19 +133,21 @@ def get_physical_topic(queue_name: str) -> SanitizedName:
 
 
 class BaseModel(pydantic.BaseModel):
-    model_config = pydantic.ConfigDict(serialize_by_alias=True)
+    model_config = pydantic.ConfigDict(
```
msullivan, line 143: Add a brief comment for why it is set up like this? (So that python construction requires snake_case but from_wire requires camelCase)

msullivan (approved):

PR #327, opened 2026-08-25
workflow: fail runs on hook conflicts

fantix:
This PR fixes a bug that, `await UserHook.wait(token=SAME_TOKEN)` may hang forever under a race condition.

> When two active runs claimed the same token, the World correctly persisted a `hook_conflict` event for the loser, but the Python handler ignored the event returned from `hook_created` and acknowledged the queue delivery. Nothing then woke the run to pass that conflict back to the workflow.

The handler now performs another replay pass inside the same delivery, after receiving the `hook_conflict` from the world, to fail the run. `BaseHook.wait()` may now raise a newly-added `vercel.workflow.HookConflictError` (new public API), carrying the conflicting `token` and an optional `conflicting_run_id`

---

## Summary

- Detect hook conflict events returned from hook creation and cold-replay the workflow immediately.
- Add a public `HookConflictError` with Python/TypeScript-compatible serialization.
- Cover conflict liveness, pending waits, typed causes, and non-conflicting registration behavior.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -990,11 +990,10 @@ class WorkflowOrchestratorContext:
                     hook = self.suspensions.pop(event.correlation_id, None)
                     if hook is not None:
                         assert isinstance(hook, Hook)
-                        conflict = f'Hook token "{token}" is already in use by another workflow'
                         while hook.futures:
                             future = hook.futures.popleft()
                             if not future.cancelled():
-                                future.set_exception(RuntimeError(conflict))
+                                future.set_exception(errors.HookConflictError(token))
 
                 case w.HookReceivedEvent(event_data=w.HookReceivedEventData(payload=data)):
                     hook = self.suspensions[event.correlation_id]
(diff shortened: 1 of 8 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1215,6 +1214,13 @@ def refuse_cross_environment_delivery(
     return True
 
 
+class _ReplayImmediately:
+    """Ask the current queue delivery to cold-replay without acknowledging it."""
```
msullivan:
```suggestion
    """Sentinel to have workflow_handler rerun _workflow_replay_pass"""
```

cold-replay is some made-up claude jargon :P

msullivan (approved):
Looks good!

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: lol yeah it was trying to say "full replay with fresh sandbox and all event logs from the start" comparing to a JS mechanism where you can resume the current replay by appending to event log queues on the fly. But we never inherited those (yet) so all replays are "cold", so it doesn't make sense to mention that. I'm pretty sure I dropped this docstring ... but yours definitely looks better!

PR #328, opened 2026-08-25
workflow: add hook conflict inspection

fantix:
This PR adds support for a workflow to check whether a hook token is already in use without waiting for someone to send data to the hook.

New public API (identical to TS): `HookEvent.get_conflict() -> Run[Any] | None`. This API registers the hook's token if it is free. It returns `None` when registration succeeds, or the owning `Run` if another hook already uses the token.

```python
hook = Approval.wait(token=f"order:{order_id}")

conflict = await hook.get_conflict()
if conflict is not None:
    return {"status": "already-running", "run_id": conflict.run_id}

approval = await hook
return {"status": "approved", "by": approval.user_id}
```

---

- Keep the owner run ID when a hook token conflicts.
- Add `HookEvent.get_conflict()` to claim a token without waiting for hook data.
- Continue the workflow as soon as that token claim succeeds.
- Return the same owning `Run` from repeated conflict checks.
- Materialize lazy Vercel hook conflict events so the runtime can inspect them.


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1068,14 +1118,33 @@ def resume(self) -> None:
...
-                            future.set_exception(errors.HookConflictError(token))
+                            future.set_exception(conflict_error)
+                    while conflicting_hook.conflict_futures:
+                        future = conflicting_hook.conflict_futures.popleft()
+                        if future.cancelled():
+                            continue
+                        if conflicting_hook.conflicting_run is not None:
+                            future.set_result(conflicting_hook.conflicting_run)
+                        else:
+                            future.set_exception(conflict_error)
```
msullivan, line 1149: I don't really understand this distinction; it probably needs a comment about it.
But what is the actual case where we might have a HookConflict without a conflicting_run_id?
Something seems a little odd about it.

msullivan (approved):

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Ah this was for pre-existing data that didn't set `conflicting_run_id` in the event, and mostly for the local world. TS has the same optional `conflictingRunId` too, ig the Postgres world also needed to be compatible across adding the conflictingRunId feature. Will add a comment here.

PR #331, opened 2026-08-26
Expand CI test coverage

elprans:
Run lint and type checking once instead of repeating them across the
Python matrix, but expand tests and distribution builds to macOS
and Windows.  Use `poe` directly instead of calling legacy scripts.


scotttrinh (approved):
Change looks good to me!

PR #332, opened 2026-08-26
workflow: prevent workflows from having side effects while suspending

msullivan:
We would *like* to just free all the coroutines in a suspended
workflow without running any more of their code, but Python doesn't
allow that (if a generator/coroutine isn't closed, it will throw a
GeneratorExit into it while destructing it).

To make sure that nothing untoward happens, add a `suspended` flag and
have all workflow operations raise CancelledError if it is set.

One big thing this fixes is `hook.dispose()` in a `finally` block.
(We could probably make HookEvent a context manager now?)

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -717,9 +717,13 @@ class WorkflowOrchestratorContext:
             except asyncio.CancelledError:
                 if not self.suspended:
                     raise RuntimeError("workflow was cancelled") from None
-                return None
             finally:
                 self._ctx.reset(token)
+                # Turn suspended into a None return regardless of what the
+                # task actually did with it.
+                # noqa because we are *intentionally* silencing an exception here.
+                if self.suspended:
+                    return None  # noqa: B012
 
     async def run_step(self, step: core.Step[P, T], *args: P.args, **kwargs: P.kwargs) -> T:
         # Bound to the step's own signature, so the recorded bytes depend on it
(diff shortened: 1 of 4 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -682,6 +714,10 @@ def run_workflow(self: Self, workflow_run: w.WorkflowRun) -> bytes:
                         )
                     )
                 )
+            except asyncio.CancelledError:
+                if not self.suspended:
+                    raise RuntimeError("workflow was cancelled") from None
+                return None
```
fantix: What if the user captured our `CancelledError` and returned or raised another error? Like:

```python
@registry.workflow
async def receive_one_suppressing_cancellation() -> int:
    hook = Ping.wait(token=CANCEL_SUPPRESSED_TOKEN)
    try:
        payload = await hook
    except asyncio.CancelledError:
        return -1
    assert payload is not None and payload.n is not None
    return payload.n
```

msullivan: Good catch! Fixed

fantix (approved):

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Aw would [PEP 765](https://peps.python.org/pep-0765) be a concern to return in `finally`?

msullivan: Oh, argh, right. I had actually remembered talking to people about that but figured it might not have made it in, if CI is passing. But they made it a warning? Sigh. I'll go add some code duplication.

PR #335, opened 2026-08-26
Guard unpublished package versions

elprans:
Treat 0.0.0 as the sentinel for packages that have not been published to
PyPI yet (aka new packages).  Skip sentinel packages during publishing
and require newly added packages to use it so merging package code
cannot publish to PyPI unintentionally.

### DIFF
--- a/tests/unit/test_release_system.py
+++ b/tests/unit/test_release_system.py
@@ -63,6 +63,72 @@ def test_compute_releases_cascades_dependency_only_patch(
     ]
 
 
+def test_compute_releases_promotes_unpublished_feature_to_initial_minor(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    changes = tmp_path / "changes"
+    fragment_dir = changes / "pkg"
+    fragment_dir.mkdir(parents=True)
+    (fragment_dir / "initial.feature.md").write_text("Add package.\n", encoding="utf-8")
+
+    version_file = tmp_path / "pkg/version.py"
+    version_file.parent.mkdir()
+    version_file.write_text('__version__ = "0.0.0"\n', encoding="utf-8")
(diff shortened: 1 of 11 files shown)


tests/unit/test_release_system.py
```diff
@@ -62,6 +62,26 @@ def test_compute_releases_cascades_dependency_only_patch(
     ]
 
 
+def test_compute_releases_promotes_unpublished_feature_to_initial_minor(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    changes = tmp_path / "changes"
+    fragment_dir = changes / "pkg"
+    fragment_dir.mkdir(parents=True)
+    (fragment_dir / "initial.feature.md").write_text("Add package.\n", encoding="utf-8")
```
scotttrinh, line 72: Does this mean that the mechanism to get your new package to be picked up and released at 0.1.0 is to add a news fragment? If so, we want to avoid adding news fragments until we're ready for the first release of the package, is that right?

elprans: I think you're required to add a news fragment with the new package.

scotttrinh: What I'm trying to figure out is: are we trying to allow development without publishing for a while, keeping the version at 0.0.0 and then triggering the first published version which will bump it to 0.1.0? Or is the workflow that you introduce a new package at 0.0.0 (even if it's broken or partial) and it is immediately published as 0.1.0 during the next release?

elprans: We can allow that if that's the behavior we want

scotttrinh: It seems like a nice-to-have. Avoids accidentally publishing something that isn't working yet. But definitely not a blocker.

scotttrinh (approved):

tests/unit/test_release_system.py
elprans: Made it so.

PR #336, opened 2026-08-26
Advance workspace dependency bounds

scotttrinh:
When a workspace package release reaches a dependent's existing exclusive upper bound, release preparation now advances that bound to the next compatibility boundary in the same generated release PR. This keeps deliberate compatibility bounds such as `vercel-internal-core<0.2.0` from producing an uninstallable release when internal core first moves to `0.2.0`.


scripts/release.py
```diff
@@ -240,6 +248,75 @@ def write_changelog(
...
+        )
+    planned = Version(version)
+    next_boundary = _next_compatibility_boundary(planned)
+    extras = f"[{','.join(sorted(requirement.extras))}]" if requirement.extras else ""
+    specifiers = [
+        f"<{next_boundary}" if item == upper_bounds[0] else str(item)
+        for item in requirement.specifier
+    ]
+    marker = f" ; {requirement.marker}" if requirement.marker else ""
+    updated = f"{requirement.name}{extras}{','.join(specifiers)}{marker}"
```
elprans, line 386: packaging normalizes the marker to double quotes, producing:

`dependencies = ["core>=0.1,<0.3.0 ; python_version < "3.11""]`

`tomllib.loads()` then raises:

`TOMLDecodeError: Unclosed array`

scripts/release.py
```diff
@@ -14,6 +14,14 @@
 from datetime import date, datetime, timezone
 from pathlib import Path
 
+from packaging.requirements import Requirement
```
elprans, line 18: Add `packaging` as an explicit dev dep.

scripts/release.py
scotttrinh: ooof, thanks for catching that. I'm going to add some (failing) e2e tests to replace a bunch of monkeypatched stub-based tests to catch real-world issues like this. Going to push it failing now, and follow-up with the simple fix.

scotttrinh: @elprans ⚡ [1a4c721](https://github.com/vercel/vercel-py/pull/336/commits/1a4c721b7abbc481fef4ebaf869620ae7ad9bcd1)

Moved a _few_ mock-based unit tests into this fixture test, so take a look there and I'm open to feedback on expanding it further, or ditching this approach and updating some mocks.

PR #337, opened 2026-08-26
Turn on check_untyped_defs, add some more annotations

msullivan:
check_untyped_defs will keep us from drowning in warnings
about type annotations and stuff in unannotated test methods.


scotttrinh (approved):
🏆

PR #339, opened 2026-08-26
workflow: fix bugs in out-of-order hook application

msullivan:
1. If there are inapplicable ooo hook events but no normal events, we
   need to suspend.
2. We need to search all ooo hooks, not just the most recent.

---


fantix (approved):
LGTM; we may eventually drop this ooo hack and use a proper event queue architecture like the TS SDK did when we have time.

PR #341, opened 2026-08-26
workflow: refactor resume() to not have a loop

msullivan:
resume() is currently structured as a loop that tries to process
events until it processes one that wakes up a suspension, at which
point it breaks. If it wakes nothing, it suspends the workflow.

When I reworked resume in #156 I left it as a loop to make the diff
more manageable (... though comments claimed it is an optimization,
which it very slightly is), but at the cost of leaving a loop with
*extremely* gnarly control flow. (It has `continue`, `break`, and
`return` all inside the loop, an *unconditional* `break` at the bottom
of the loop, and an `else` branch.)

Get rid of the loop and reduce a bunch of nesting depth.  Now the
function always resolves either zero or one event and suspends if it
resolved zero. The one callsite is already called inside a loop that
has a check for runnable tasks, so we don't need to embed that logic
ourselves.

I had previously written a bunch of slop tests for resume which
started failing even though the new version still obeys the documented
API of the function.

The updated tests work with the old version too. They might still be
slop that isn't worth having though?

### DIFF
--- a/changes/vercel-workflow/refactored-event-replay.internal.md
+++ b/changes/vercel-workflow/refactored-event-replay.internal.md
@@ -1,0 +1,1 @@
+Refactored event replay.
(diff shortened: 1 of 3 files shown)


fantix (approved):
:+1:

PR #343, opened 2026-08-27
workflow: implement call_later/call_at in custom event loop

msullivan:
The implementation is built on top of run_wait in our workflow core;
we just needed to hook up the future from that to an
asyncio.TimeHandle.

run_wait() goes from being a coroutine to being a regular function
that returns a future, so that the loop can directly install a
callback on it without needing to create a task.
The other run_*() methods get a similar treatment for consistency.
(It's probably a very small microoptimization, too.)

This unlocks `asyncio.sleep` but probably more importantly it unlocks
all of asyncio's timeout features.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/loop.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/loop.py
@@ -1,8 +1,11 @@
 import asyncio
 import contextvars
+import datetime
 import sys
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Protocol
+
+from vercel._internal.core.polyfills import UTC
 
 if sys.version_info >= (3, 11):
     from typing import TypeVarTuple, Unpack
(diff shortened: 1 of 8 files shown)


src/vercel-workflow/vercel/workflow/_internal/loop.py
```diff
@@ -32,22 +46,60 @@ def _run_once(self) -> None:
...
+            self._timers.pop(timer, None)
+            if fut.cancelled():
+                pass
+            elif fut.exception():
+                self.call_exception_handler(
+                    {
+                        "message": "an error occurred waiting for a workflow sleep",
+                        "exception": fut.exception(),
+                    }
+                )
```
fantix, line 77: Seems like we also need to terminate the loop in the next iteration (through some internal fatal flag and/or remembering the exception itself?) instead of just reporting. Re `runtime.py`:

https://github.com/vercel/vercel-py/blob/a366656e1f1da71816480b5c70bcb953af2407b5/src/vercel-workflow/vercel/workflow/_internal/runtime.py#L916-L931

This early return expects the suspension itself to terminate the loop, but `TimerHandle` cannot do that just by itself. Then, we ended up in a busy loop (agent-produced repro script):

https://gist.github.com/fantix/34dada00f7e27fdc538eaf2120f26598

fantix (approved):
LGTM other than a marginal issue on a not-so-happy path.

src/vercel-workflow/vercel/workflow/_internal/loop.py
msullivan: I think that I will make it so that we *for now* still run the callback on an error, so the failiure message will print but not hang, and then I'll merge this.

And then I'll do a follow-up cleaning up this edge case? One big question here is: currently user code can intercept and swallow or process NondeterminismErrors. Do we want to eliminate that? Probably? But we might still want to *try* signalling it via user-code, because it gives a better traceback.

So the algorithm might be like:
* Fail the suspension, and also save the NondeterminismError locally
* Suspend the workflow so everything will certainly die
* If there is a NondeterminismError saved, prioritize that over treating it as a suspension?

fantix: Yeah that sounds like a good plan! User-space traceback is great to keep.

PR #345, opened 2026-08-27
sandbox: Align create_process(stdout=None) to stdlib

scotttrinh:
Match `Popen` and `run_process` by inheriting output by default. Accept text streams so callers can stream directly without manually consuming readers. Drain the log transport before reporting completion so inherited streams receive all output and `PIPE` readers reach EOF reliably.


fantix (approved):
LGTM for the fix itself, but it surfaced a new divergence to `Popen`:

After `proc = Popen(..., stdout=None)`, the output would already start to flush stdout (or whatever inherited stream); but `proc = await create_process(stdout=None)` would wait until `proc.wait()` or `proc.communiate()` is called to drain the remote output.

elprans (approved):
Make sure you've double-checked all tests and examples.

scotttrinh:
Going to actually abandon this since it's pretty hard to achieve a symmetrical API between sync and async here that doesn't do a bunch of implicit stuff you wouldn't expect. The current behavior is already pretty intuitive and having to explicitly print to stdout seems like a better API (for now at least) then the other ideas we came up with. Thanks @fantix for helping to talk me through the design here.

PR #347, opened 2026-08-28
workflow: fail more reliably on nondeterminism errors

msullivan:
We need to make sure that a nondeterminism error terminates the run
promptly, even if the application swallows it or is not directly
waiting on it.

Do this by:
 - Stashing the error in the workflow context
 - Suspending the task
 - Raising the error as a top priority once the workflow task stops
 - Some extra cleverness to preserve the real traceback in the common
   case where the exception *did* get raised without trouble.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -659,6 +659,9 @@ class WorkflowOrchestratorContext:
         self.registry = registry
 
         self.suspended = False
+        # An error the run must fail with regardless of what the body does;
+        # raised by run_workflow(). See _fail_nondeterminism.
+        self.resume_exception: Exception | None = None
 
     @classmethod
     def current(cls) -> Self:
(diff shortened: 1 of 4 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -741,6 +757,12 @@ def run_workflow(self: Self, workflow_run: w.WorkflowRun) -> bytes | None:
                     )
                 )
             except BaseException as ex:
+                if self.resume_exception is not None:
+                    # If the body surfaced the error itself, keep its
+                    # traceback.
+                    if type(ex) is type(self.resume_exception):
+                        raise
```
fantix: Reading the comments, I was expecting `if ex is self.resume_exception:`. Then I saw the test and it makes sense now. Not a change request, just stating my confusion.

fantix (approved):

src/vercel-workflow/vercel/workflow/_internal/runtime.py
msullivan: Actually, looking at this, I could not totally reconstruct why I had told the agent that the check needed to be based on type. I think my intuition was that the exception would be recreated, but that's not true! There is a test for a reraise of something of the same type, but that was slop I don't care about.

And thinking about it more (and testing), I don't think I even need the `ex is self.resume_exception` at all! The traceback will already be on self.resume_exception since it got threaded up.

msullivan: And there's a possible benefit to dropping the check entirely: even if it didn't make it out, resume_exception will have picked up a traceback anyway!

msullivan: > And there's a possible benefit to dropping the check entirely: even if it didn't make it out, resume_exception will have picked up a traceback anyway!

Oh actually we had that anyway. Check was pure dead code, accomplished nothing, damaged nothing.

PR #348, opened 2026-08-28
workflow: add opt-in cancellable steps

msullivan:
Enable with `@workflows.step(cancellable=True)`.

When `cancel()` is called on a cancellable step, we send a message on
a stream that the step will listen for. If it gets a message, it will
exit. It gets recorded in the event log as a hook read. (This makes
more sense implementing JS AbortController than it does here, but we
need to record it somehow.)

Note that like regular asyncio tasks, `cancel()` does not cause the
step to immediately become "cancelled". Waiting on it will still wait
for the step to actually terminate.

If there are still cancellable steps running when a workflow function
completes, they will be cancelled and the workflow will wait for them
to finish before terminating.

I am unsure about that particular part of the design. It could be
potentially useful sometimes to leave them running, also.  This
approach best leverages what `asyncio.run` does naturally with tasks,
which is how everything else gets run here.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -138,8 +138,7 @@ class Cancellation(BaseSuspension):
     requested: bool = False
 
     def fail(self, exc: Exception) -> None:
-        # Cancellation suspensions have no direct awaiter. The orchestrator
-        # raises replay-divergence failures separately via ``resume_exception``.
+        # Cancellation suspensions have no awaiter.
         pass
 
 
@@ -1149,12 +1148,14 @@ class WorkflowOrchestratorContext:
 
             case w.HookCreatedEvent():
                 hook = self.suspensions[event.correlation_id]
-                assert isinstance(hook, Hook)
                 hook.has_created_event = True
-                while hook.conflict_futures:
-                    future = hook.conflict_futures.popleft()
-                    if not future.cancelled():
-                        future.set_result(None)
+                if isinstance(hook, Hook):
+                    while hook.conflict_futures:
+                        future = hook.conflict_futures.popleft()
+                        if not future.cancelled():
+                            future.set_result(None)
+                else:
+                    assert isinstance(hook, Cancellation)
 
             case w.WaitCreatedEvent():
                 self.suspensions[event.correlation_id].has_created_event = True
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -131,14 +131,16 @@ class Cancellation(BaseSuspension):
     """
 
     token: str
-    step: Suspension[Any]
+    step_id: str
     reason: str | None = None
     # Has a cancellation been requested ever? Prevents repeat cancellations
     # from being sent.
     requested: bool = False
 
     def fail(self, exc: Exception) -> None:
-        self.step.fail(exc)
+        # Cancellation suspensions have no direct awaiter. The orchestrator
+        # raises replay-divergence failures separately via ``resume_exception``.
+        pass
 
 
 class CallbackCancelFuture(asyncio.Future[T]):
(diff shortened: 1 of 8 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -791,8 +846,40 @@ def run_step(
         input_data = ser.dehydrate(ser.step_arguments(dumped_args, dumped_kwargs))
         sus = Suspension(correlation_id=f"step_{self.generate_ulid()}", step=step, input=input_data)
         self.suspensions[sus.correlation_id] = sus
+        if step.cancellable:
+            cancellation = Cancellation(
+                correlation_id=f"hook_{self.generate_ulid()}",
+                token=f"abrt_{_correlation_ulid(sus.correlation_id)}",
```
fantix: Extract a local ulid variable rather than parsing the just-generated `correlation_id`?

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -791,8 +846,40 @@ def run_step(
...
 
+    def _on_step_cancel(self, cancellation: Cancellation, msg: object) -> bool:
+        """Handle cancel() of a cancellable step's future; see CallbackCancelFuture.
+
+        When cancel() is called during normal execution, we install a
+        cancellation suspension for the step, which will get picked up
+        when the run suspends and turned into a cancellation.
+
+        We suppress the *actual* cancellation of the future, because
+        we anything waiting on the step to still wait until it has
```
fantix: typo: we **want** (or need?) anything ... to still wait until ...

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -791,8 +846,40 @@ def run_step(
...
+
+        If the loop is suspended, though, we aren't doing side effects
+        and we just want to tear things down, so allow a normal cancel.
+
+        """
+        if self.suspended:
+            return True
+        if not cancellation.requested:
+            cancellation.requested = True
+            cancellation.reason = str(msg) if msg else None
```
fantix: `if msg` should be `if msg is not None`? I think asyncio `fut.cancel("")` carries over the falsy empty string as a valid message, and maybe we want to be consistent with asyncio.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1483,6 +1630,9 @@ async def _workflow_replay_pass(
     except Exception as e:
         error_message = "".join(traceback.format_exception_only(type(e), e)).strip()
         logger.exception("[Workflows] '%s' - workflow run failed: %s", run_id, error_message)
+        # Before the terminal event, which would reject them; a failure here
+        # fails the delivery, so the whole pass is retried.
```
fantix:
```suggestion
        # Flush the step cancellation events before RunFailed, because a terminal
        # event would reject event creation afterwards, and we also need the
        # workflow run to retry if the flush itself fails.
```

Or drop the comment idk because the code is already clear by itself?

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1496,6 +1646,7 @@ async def _workflow_replay_pass(
         return None
 
     if output is not None:
+        await _send_cancellations(context)
```
fantix:
```suggestion
    await _send_cancellations(context)

    if output is not None:
```

So that we can also drop the same call at line 1659.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -108,6 +108,52 @@ class Suspension(BaseSuspension, Generic[T]):
...
+class Cancellation(BaseSuspension):
+    """The pending cancellation of a cancellable step.
+
+    Registered in the run's suspensions when the body first cancels the step;
+    removed when the recorded ``hook_received`` replays, so a still-registered
+    one at flush time is exactly a cancellation left to send.
+    """
+
+    token: str
+    step: Suspension[Any]
```
fantix: Looks like we only used `step.correlation_id`, maybe we can store only that instead.

fantix (approved):
Cool!

PR #353, opened 2026-08-28
workflow: make `await hook` never return `None`

msullivan:
Technically a breaking change, as now it can raise `HookDisposedError`.

Disposing a hook while waiting on it will not be a common enough
pattern to have it be worth returning `| None` from the hook.

(In TS, the hook just hangs forever in that case.)


fantix (approved):
LGTM; though we may need to update the protocol too, like what `HOOK_CONFLICT_ERROR_TAG` does, so that the client caller can pick it up through `run.return_value()` and `WorkflowRunFailedError.error`. In that case, we'd also need the TS-parity of `HookDisposedError` in the future... I now feel `HookDisposedError` is more like a local exception that should be handled within the workflow and shouldn't cross the serialization border.

msullivan:
> LGTM; though we may need to update the protocol too, like what `HOOK_CONFLICT_ERROR_TAG` does, so that the client caller can pick it up through `run.return_value()` and `WorkflowRunFailedError.error`. In that case, we'd also need the TS-parity of `HookDisposedError` in the future... I now feel `HookDisposedError` is more like a local exception that should be handled within the workflow and shouldn't cross the serialization border.

My though is that it's an application-level issue that should get surfaced to the application, which is what this does. I'm not totally sure how we'd handle it within the workflow?

PR #354, opened 2026-08-28
workflow: make `HookEvent` an async context manager

msullivan:
This matches TS, which supports `using`, and is also idiomatic.
```
async with SomeHook.wait(...) as hook:
    res = await hook
```

The one open question is whether it should be a sync context manager
or an async one. It *can* be a sync one (dispose() is sync), but if we
think it is going to be common to mix it with async managers, then
making it async is more convenient.

### DIFF
--- a/changes/vercel-workflow/hook-context-manager.feature.md
+++ b/changes/vercel-workflow/hook-context-manager.feature.md
@@ -1,0 +1,8 @@
+Make `HookEvent` an async context manager
+
+This matches TS, which supports `using`.
+```
+# disposes the hook on block exit
+async with SomeHook.wait(...) as hook:
+    res = await hook
+```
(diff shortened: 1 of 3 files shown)


fantix (approved):
> The one open question is whether it should be a sync context manager or an async one. It *can* be a sync one (dispose() is sync), but if we think it is going to be common to mix it with async managers, then making it async is more convenient.

Would it be a bad idea to have both sync and async context managers?

msullivan:
> > The one open question is whether it should be a sync context manager or an async one. It _can_ be a sync one (dispose() is sync), but if we think it is going to be common to mix it with async managers, then making it async is more convenient.
> 
> Would it be a bad idea to have both sync and async context managers?

I've done this before in other places and Yury made me take it out :P

PR #355, opened 2026-08-28
workflow: Improve portability

elprans:
Avoid import finder recursion by returning already-loaded sandbox
imports directly and performing third-party spec discovery in the host
import context.

Keep platform behavior host-native and decode Node CLI output as UTF-8
so the Windows test matrix passes.


msullivan (changes requested):
Would it be possible to get the tests passing on windows without making it pretend to be posix? Then we can decide separately from bugfixes whether we need/want to do that?

elprans:
> Would it be possible to get the tests passing on windows without making it pretend to be posix? Then we can decide separately from bugfixes whether we need/want to do that?

Made it so.

fantix (approved):

msullivan (approved):

PR #356, opened 2026-08-28
Port to httpx2

elprans:
Revive the httpx2 migration from #150.

Because respx upstream does not support httpx2 yet we vendor respx 0.23.1
and search-and-replace port it to httpx2.  The alternative is runtime
monkeypatching and typing stubs, neither of which is particularly
pleasant.


msullivan (approved):

changes/vercel-sandbox/httpx2.feature.md
```diff
@@ -0,0 +1 @@
+Replace the unmaintained `httpx` dependency with its maintained `httpx2` successor.
```
scotttrinh: I think this is breaking for sandbox unless you are still able to pass `httpx` in the client factory.

fantix: Likewise for `vercel-queue`:

https://github.com/vercel/vercel-py/blob/44dbf1e0533a23ea4fb91c74c4ebcb88ab1dc0d6/src/vercel-queue/README.md?plain=1#L315

But I think this is not entirely true now since we are e.g. capturing `httpx2.TransportError` and reraising `vercel.queue.CommunicationError`; passing in an old `httpx` client factory would break the raised error.

scotttrinh: Maybe it makes sense to catch both library's errors for now? I'm also ~fine with this being a breaking change, we're moving fast right now and we probably don't want to slow down just yet.

elprans: Added explicit compat.  Users can still pass httpx clients for now.

PR #362, opened 2026-09-02
workflow: tolerate closed cancellation streams

fantix:
On `VercelWorld`, concurrent workflow replays may try to cancel the same step at the same time. Once one replay sends the cancellation and closes the abort stream, the other gets HTTP 409. It should ignore that response and continue recording the cancellation events.

This was caught by the E2E test added in vercel/workflow#3930.
Refs #348.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -1560,7 +1560,10 @@ async def _send_cancellation(world: w.World, run_id: str, cancellation: Cancella
         await world.streams_write(run_id, stream_name, streams.encode_frame(payload))
         await world.streams_close(run_id, stream_name)
     except w.WorkflowWorldError as error:
-        if error.status not in (400, 409):
+        # Concurrent replays can both try to send the same cancellation. Once
+        # one replay closes the stream, Vercel rejects the other's write with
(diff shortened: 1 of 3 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1560,7 +1560,7 @@ async def _send_cancellation(world: w.World, run_id: str, cancellation: Cancella
         await world.streams_write(run_id, stream_name, streams.encode_frame(payload))
         await world.streams_close(run_id, stream_name)
     except w.WorkflowWorldError as error:
-        if error.status != 400:
+        if error.status not in (400, 409):
```
msullivan: Let's comment what the errors we are OK with are?
(I know that I did not follow this when I produced this original line of code though!)

msullivan (approved):

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: My agent searched the server code and warned me about recovering from 400: it may hide a real bug in bad requests rather than continuing on conflict. What was your use case recovering from the 400 error? I may drop 400 and keep 409 if it was meant for the same scenario?

PR #369, opened 2026-09-02
workflow: read compressed payloads

fantix:
This PR reads gzip- and zstd-compressed workflow payload envelopes, and aligned the hydration with the TypeScript decrypt, decompress, then deserialize pipeline.

Added new dependency `zstandard` on Python <3.14.


src/vercel-workflow/vercel/workflow/_internal/serialization.py
```diff
@@ -162,15 +165,7 @@ def hydrate(data: Any, *, what: str, key: bytes | None = None) -> Any:
     if not isinstance(data, bytes | bytearray | memoryview):
         raise SerializationError(f"{what} is not serialized data: {type(data).__name__}")
     data = bytes(data)
```
fantix, line 167: should probably use zerocopy `memoryview` - in a new PR.

msullivan (approved):

PR #373, opened 2026-09-03
refactor: extract `wait_until` context from cache

fantix:
Upcoming workflow `ReadableStream` would start background tasks to pump data from user generators into a workflow stream. Such task should continue to run to completion even after the step handler returns. This is what `vercel.functions.wait_until(awaitable)` offers, and the Workflow JS SDK is using similar mechanism.

The problem is, neither `vercel.functions` (the `vercel` package) nor the underlying context storage of `wait_until` (currently stored in `vercel.cache.context`, in `vercel-cache` package) can be imported from `vercel-workflow` - workflow doesn't have to depend on cache.

This PR moves that context storage to `vercel.functions.context` in `vercel-internal-core`. The rest of `vercel.functions` stays in the `vercel` meta package, with `pkgutil.extend_path()` allowing the workspace to combine it with the context subpackage from internal core.

`vercel.functions.context` is not considered public API - it's only used by the runtime and other `vercel-*` packages internally. Therefore, `vercel.functions` didn't re-export what's in the context module.

This also allows the runtime to avoid using the confusing `vercel.cache.context.set_context()`, but rather use the more accurate `vercel.functions.context.set_wait_until()`. See also the paired runtime change: https://github.com/vercel/vercel-internal/pull/786

See also: #2, #59, #149, #218


fantix:
This is decided in a meeting to get moved into `vercel-context` package.

PR #374, opened 2026-09-03
workflow: write encrypted payloads

fantix:
Encrypt workflow payload writes and seal external hook resumes to the run's public key.

Streamed data is not encrypted yet.

Also publishes a synthetic workflowCoreVersion, which is apparently used (improperly) as capability detection in the JS SDK.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/encryption.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/encryption.py
@@ -20,7 +20,7 @@ import base64
 import binascii
 import hashlib
 import hmac
-import os
+import secrets
 from typing import Any, NamedTuple
 
 from cryptography.exceptions import InvalidTag
@@ -59,7 +59,7 @@ def encrypt_envelope(key: bytes, plaintext: bytes) -> bytes:
     """Seal *plaintext* with AES-256-GCM, returning ``nonce + ciphertext + tag``."""
     if len(key) != KEY_LENGTH:
         raise ValueError(f"A run key is {KEY_LENGTH} bytes, got {len(key)}")
-    nonce = os.urandom(NONCE_LENGTH)
+    nonce = secrets.token_bytes(NONCE_LENGTH)
     return nonce + AESGCM(key).encrypt(nonce, plaintext, None)
 
 
@@ -252,7 +252,7 @@ def seal_envelope(recipient_public_key: bytes, plaintext: bytes) -> bytes:
     """
     if len(recipient_public_key) != KEY_LENGTH:
         raise ValueError(f"A run public key is {KEY_LENGTH} bytes, got {len(recipient_public_key)}")
-    ephemeral_scalar = os.urandom(KEY_LENGTH)
+    ephemeral_scalar = secrets.token_bytes(KEY_LENGTH)
     ephemeral_public_key = _x25519_public_key(ephemeral_scalar)
     shared = _x25519_exchange(ephemeral_scalar, recipient_public_key)
     if shared is None:
--- a/src/vercel-workflow/vercel/workflow/_internal/encryption.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/encryption.py
@@ -1,7 +1,4 @@
-"""The `encr` and `encp` payload formats, as `@workflow/world-vercel` writes them.
-
-Decryption only, for now: writing either is not implemented, so the payloads
-this SDK produces are unencrypted `devl`, `gzip`, or `zstd` envelopes.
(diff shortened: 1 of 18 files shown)


src/vercel-workflow/vercel/workflow/_internal/encryption.py
```diff
@@ -57,6 +55,14 @@ def _aes_gcm_decrypt(key: bytes, nonce: bytes, sealed: bytes) -> bytes | None:
         return None
 
 
+def encrypt_envelope(key: bytes, plaintext: bytes) -> bytes:
+    """Seal *plaintext* with AES-256-GCM, returning ``nonce + ciphertext + tag``."""
+    if len(key) != KEY_LENGTH:
+        raise ValueError(f"A run key is {KEY_LENGTH} bytes, got {len(key)}")
+    nonce = os.urandom(NONCE_LENGTH)
```
msullivan: I think it might be very marginally more tasteful to do `secrets.token_bytes(NONCE_LENGTH)`? (I'm pretty sure that will just call os.urandom though)

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -2660,7 +2710,23 @@ async def resume_hook(token_or_hook: str | core.Hook, payload: Any) -> core.Hook
...
+        try:
+            recipient_public_key = encryption.decode_key(
+                run.encryption_public_key, what="the run encryption public key"
+            )
+        except ValueError:
+            # Older or corrupted rows must retain the symmetric fallback.
+            pass
+    encryption_key = None
+    if recipient_public_key is None:
+        encryption_key = await world.run_key(run.run_id, deployment_id=run.deployment_id)
```
msullivan, line 2846: Maybe put this in a helper above, with a bit more of a comment?

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1955,7 +1983,9 @@ async def listen() -> None:
             async for chunk in streams.reconnecting_frames(world, run_id, stream_name):
                 cancelled = True
                 try:
-                    data = ser.hydrate(chunk, what=f"the cancellation of step {step_id}")
+                    data = ser.hydrate(
+                        chunk, what=f"the cancellation of step {step_id}", key=run_key
+                    )
```
msullivan: This is my fault but actually I think _run_cancellable_step shouldn't be using reconnecting_frames at all, really? It should use the higher level one that already does the hydration

msullivan (approved):

src/vercel-workflow/vercel/workflow/_internal/runtime.py
fantix: Yeah, ~extracted a shared function for 3 usages.~ Ahh you already extracted a function in #375, will just use that instead.

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -2040,18 +2040,21 @@ async def _run_cancellable_step(
...
-            async for chunk in streams.reconnecting_frames(world, run_id, stream_name):
-                cancelled = True
-                try:
-                    data = ser.hydrate(
-                        chunk, what=f"the cancellation of step {step_id}", key=run_key
-                    )
-                except ser.SerializationError:
-                    pass
-                else:
+            source = _read_stream(
```
msullivan, line 2043: oh I was hoping with the higher level API we also wouldn't need to thread the run_key in

src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -2712,15 +2715,31 @@ def read_stream(
...
             what = f"chunk {index} of stream {name}"
-            value = ser.hydrate(payload, what=what)
+            try:
+                value = ser.hydrate(payload, what=what, key=key)
+            except ser.SerializationError:
+                if ignore_payload_errors:
+                    # This is only used by cancellation streams: the payload carries
+                    # an optional cancel reason; the cancel happens anyway even if
+                    # the reason cannot be read.
+                    value = None
```
msullivan, line 2740: Oh huh; I know this was maintaining behavior I had added but I'm not sure being able to survive hydration failure there is worthwhile

fantix: This was also my fault producing `SerializationError` from e.g. exhausted reconnect attempts, or we could just capture `SerializationError` in `_run_cancellable_step` and remove this `ignore_payload_errors`. I will follow up with a error type cleaning PR and drop `ignore_payload_errors`.

PR #376, opened 2026-09-04
workflow: make zstandard passthrough (and restricted) to make httpx work

msullivan:
zstandard is now a dep of vercel-workflow, which made importing httpx (and thus vercel's `ai`) fail in workflows. (*Using* httpx of course *should* fail, but importing it should be OK!)

zstandard does some harmless platform checking stuff. Add it to passthrough but blocklist its `open` method.


fantix (approved):
The `_is_passthrough` and `_host` fix makes sense!

PR #378, opened 2026-09-04
workflow: rework py-sandboxing of asyncio

msullivan:
* The asyncio module was not having restrictions applied to it, so a bunch of state modification functions were allowed.
* We don't need to *proxy* the event loop any more, because we already are running an actually inherently restricted event loop.
* But move more principled blocking into the loop.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
@@ -13,7 +13,6 @@ import sys
 import threading
 import types
 import typing
-import weakref
 from collections.abc import Callable, Iterator, Mapping, MutableMapping, Sequence
 from contextlib import contextmanager
 from importlib.abc import Loader, MetaPathFinder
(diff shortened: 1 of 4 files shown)


src/vercel-workflow/vercel/workflow/_internal/py_sandbox.py
```diff
@@ -370,6 +259,40 @@ def _host_system() -> str:
...
+        "create_subprocess_shell",
+        # Network / socket I/O
+        "open_connection",
+        "open_unix_connection",
+        "start_server",
+        "start_unix_server",
+        # Bridging external threaded futures
+        "wrap_future",
+    ),
+    "asyncio.events": _blocklist(
```
fantix: I think `asyncio.events` is also re-exported on the `asyncio` module itself, so

```
import asyncio
asyncio.events.new_event_loop()
```

would work inside the sandbox, which is an escape.

fantix (approved):
LGTM other than a marginal escape.

PR #380, opened 2026-09-05
workflow: create hooks on first suspend after `.wait()`

msullivan:
This is similar to how `dispose` operates, where it is triggered without its own await. It matches JS.

This is accomplished just by registering it in suspensions in `create_hook`.

We then never remove it until it is disposed, and we decide whether to put something in the ooo queue based on whether there are waiters, not based on its presence in suspensions. This fixes an existing bug where messages that come in between `get_conflict()` and an actual wait on the hook would get dropped.

Also, adjust even creation so that the actual step queueing is done *after* all events are created. This ensures that, if we create a hook and then publish its token with a step, the hook will be created before the step executes.

### DIFF
--- a/src/vercel-workflow/vercel/workflow/_internal/runtime.py
+++ b/src/vercel-workflow/vercel/workflow/_internal/runtime.py
@@ -13,7 +13,15 @@ import random
 import sys
 import traceback
 from collections import deque
-from collections.abc import AsyncGenerator, AsyncIterator, Callable, Coroutine, Mapping, Sequence
+from collections.abc import (
+    AsyncGenerator,
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Coroutine,
+    Mapping,
+    Sequence,
+)
 from datetime import datetime
 from typing import Any, Generic, Literal, ParamSpec, TypeVar, cast, overload
 from urllib.parse import parse_qsl, urlsplit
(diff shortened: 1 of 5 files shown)


src/vercel-workflow/vercel/workflow/_internal/runtime.py
```diff
@@ -1894,15 +1920,18 @@ async def create_step(s=sus):
...
+                        functools.partial(
+                            world.queue,
+                            w.get_queue_name(workflow_run.workflow_name, namespace),
+                            w.WorkflowInvokePayload(
+                                run_id=run_id,
+                                step_id=s.correlation_id,
+                                step_name=s.step.name,
+                                requested_at=datetime.now(UTC),
+                            ),
+                            idempotency_key=s.correlation_id,
```
msullivan, line 1933: anyio added tg.create_task that would take a coroutine in like 4.13.0 or something and we only depend on >=4.0.0, so it needs to be a callable, hence the functools.partial.

We could consider depending on a later version.

fantix: I think my agent did something similar recently - maybe that's a sign to upgrade anyio. However, `tg.create_task` was added in [4.14](https://github.com/agronholm/anyio/releases/tag/4.14.0), which was only a little less than 3 months ago, so maybe it's still better we stay with partials for a bit longer!

fantix (approved):

PR #381, opened 2026-09-05
workflow: get rid of the central ooo buffer for hook received

msullivan:
Instead, now that we are keeping Hooks alive forever, stash them in the Hook object.

I think just last week I made run_hook and similar return Futures, for consistency with run_wait, which kind of has to now. This changes run_hook back to being `async` so that it is easier for it to return in some cases and wait on the future in others.

(run_hook_conflict was already doing that by creating temporary futures, and I changed that also.)

### DIFF
--- a/changes/vercel-workflow/no-ooo-buffer.internal.md
+++ b/changes/vercel-workflow/no-ooo-buffer.internal.md
@@ -1,0 +1,1 @@
+Get rid of `ooo_hook_received_events`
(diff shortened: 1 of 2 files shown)


msullivan:
One note is that this can change the order of evaluation of workflows: previously an ooo hook could only get applied in `resume`, when everything is quiescent. Now they get registered immediately and something trying to wait on one will never block on a future, it will just get it.

fantix (approved):
LGTM!

PR #383, opened 2026-09-08
workflow: deliver hook payload errors to awaiters

fantix:
Deliver hook payload deserialization and validation errors to the hook awaiter so workflows can catch them.

~Also add a shared `Outcome[T] = Value[T] | Error` container to vercel-internal-core~

Refs #381

### DIFF
--- a/changes/vercel-internal-core/outcome.internal.md
+++ b/changes/vercel-internal-core/outcome.internal.md
@@ -1,0 +1,1 @@
+Add shared success and error outcome types.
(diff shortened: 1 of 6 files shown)


changes/vercel-internal-core/outcome.internal.md
```diff
@@ -0,0 +1 @@
+Add shared success and error outcome types.
```
msullivan: I don't feel strongly here, because it's obviously pretty marginal, but maybe we want to just keep this in workflow?

Dealing with Futures is typically best avoided, and most of the rest of the library uses `anyio` anyway, I think?

msullivan (approved):

PR #384, opened 2026-09-09
env: Move `vercel.env` to it's own package

scotttrinh:
There was a cyclical dependency between `vercel-sandbox` and `vercel` via `vercel.env`, so extract to it's own package. ~Also added a bunch of release-side hardening to ensure that we detect this in the future. Refactor the publishing workflow for more robust publishing detection, and narrow the write-enabled part of the publishing job.~

Edit: We decided to make this PR just the pure extraction, and we'll move tooling and configuration into a separate PR.


.github/scripts/test_installed_wheel.sh
```diff
@@ -27,9 +27,15 @@ for dependency_wheel in "$wheel_dir"/vercel_*.whl; do
     if [ ! -f "$dependency_wheel" ] || [ "$dependency_wheel" = "$wheel_path" ]; then
         continue
     fi
-    case "$(basename "$dependency_wheel")" in
+    base_name=$(basename "$dependency_wheel")
+    case "$base_name" in
         *_bundle-*.whl|vercel_internal_shared_vendored_deps-*.whl) ;;
-        *) continue ;;
+        *)
+            prefix=$(echo "$base_name" | sed -E 's/-[0-9].*//')
```
elprans: What is going on here?

integrations/vercel-apscheduler/pyproject.toml
```diff
@@ -20,6 +20,11 @@ dependencies = [
 vercel-cache = { workspace = true }
 vercel-queue = { workspace = true }
 
+[tool.vercel.release.bundle]
```
elprans: Use `tool.vercel-sdk` to avoid polluting the `tool.vercel` namespace which is for platform config.

.github/scripts/test_installed_wheel.sh
scotttrinh: This skips looking for `vercel_env_bundle-0.1.0.whl` which is what `vercel_sandbox_bundle-0.5.0.whl` is looking for. Looking at some other cases we have though, it seems like we just publish duplicate bundled packages when they are dependencies of another bundled package, so I'll just do that and revert this change.

integrations/vercel-apscheduler/pyproject.toml
scotttrinh: As discussed, we'll do this additional work separately, potentially in a separate config file.

elprans (approved):

PR #389, opened 2026-09-11
internal: Publish independent packages after failures

scotttrinh:
Publish in dependency order without failing fast. Block affected dependents, preserve successful tags and releases, and report package failures in Actions.

We considered generated per-package workflow jobs and a CI-generated matrix. Separate jobs required generated YAML and artifact handoffs. A matrix cannot express dependencies between its entries without staged waves or custom coordination. Keep the best-effort loop in the script and report one overall failure instead.


.github/workflows/publish.yml
```diff
@@ -39,57 +42,7 @@ jobs:
         env:
           BASE_REF: ${{ github.event.before || 'HEAD^' }}
           FORCE: ${{ github.event_name == 'workflow_dispatch' && inputs.force }}
-        run: |
```
scotttrinh, line 42: Most of this moved into publish.py

.github/workflows/publish.yml
```diff
@@ -114,216 +67,101 @@ jobs:
         env:
           PACKAGES_JSON: ${{ needs.detect.outputs.packages }}
           DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run }}
-          FORCE: ${{ github.event_name == 'workflow_dispatch' && inputs.force }}
-        run: |
```
scotttrinh, line 118: ditto for this

.github/workflows/publish.yml
```diff
@@ -114,216 +67,101 @@ jobs:
...
-            echo "EOF"
-          } >> "$GITHUB_OUTPUT"
+          PUBLISH_SHARED: ${{ needs.detect.outputs.publish-shared }}
+          SHARED_ERROR: ${{ needs.detect.outputs.shared-error }}
+          VERCEL_INTERNAL_SHARED_VENDORED_DEPS_VERSION: ${{ needs.detect.outputs.shared-version }}
+        run: python scripts/publish.py run
 
       - name: Create Git Tags
-        if: steps.publish.outputs.tags != '' && !(github.event_name == 'workflow_dispatch' && inputs.dry_run)
+        if: ${{ !cancelled() && steps.publish.outputs.tags != '' && !(github.event_name == 'workflow_dispatch' && inputs.dry_run) }}
```
scotttrinh: We want to create tags and releases for the successfully published packages still, so only skip on cancellation. This keeps the job running even if earlier steps fail.

scripts/publish.py
```diff
@@ -0,0 +1,225 @@
...
+    for name, dependencies in ordered.items():
+        if name not in selected:
+            continue
+        unavailable = set(dependencies) & (failed | blocked)
+        if unavailable:
+            blocked.add(name)
+            reason = f"Blocked by {', '.join(sorted(unavailable))}"
+            rows.append(f"| {name} | blocked | {reason} |")
+            print(f"{name}: {reason}", flush=True)
+            continue
```
scotttrinh: This is the heart of the change: if a package depends on a failed or blocked package, it's also blocked, but other packages should be attempted still rather than failing fast.

fantix (approved):
OK for this one; I'm more concerned about build failures, which usually indicates further code fixes. Maybe we should separate the build phase from publish phase, so that any build failure prevents entering publish phase for all packages.

tests/unit/test_publish.py
```diff
@@ -0,0 +1,399 @@
...
+        publish.ancestors({"a": {"b"}, "b": {"a"}})
+    with pytest.raises(ValueError, match="Unknown publish dependencies: missing"):
+        publish.ancestors({"app": {"missing"}})
+
+
+def test_shared_dependency_only_gates_its_consumers() -> None:
+    graph = publish.dependency_graph(workspace.packages())
+    shared = publish.SHARED
+    assert not graph[shared]
+    assert not graph["vercel-env"]
```
fantix: KeyError

fantix (approved):
LGTM! We can cut the `publish` job into `build` + `publish` jobs in the future, so that 1) we can see which job failed, 2) built artifacts are also available on GH action run for debugging purposes, 3) we can rerun the publish job separately reusing artifacts without having to rebuild again. At that time, perhaps we don't need the `publish/action.yml` abstraction, given that most logic is already moved into the Python script.

scotttrinh:
OK, I did a bit more refactoring and spent more time reviewing and digging in here. I have some follow up that I want to discuss more broadly, but I'm not going to try to land it in this PR.

---

**Follow-up: run authenticated release rehearsals through `publish.yml`.** This PR’s dry run builds and verifies artifacts, reads PyPI’s registry state, and checks artifact selection, but skips `uv publish` entirely. We can strengthen it by invoking `uv publish --dry-run --trusted-publishing always <artifact-files>`. uv obtains Trusted Publishing credentials before skipping uploads, exercising the GitHub OIDC → PyPI token exchange without creating a release. Its dry-run path also validates artifact metadata; `--check-url https://pypi.org/simple/` adds existing-file hash comparisons. See [uv’s publishing implementation](https://github.com/astral-sh/uv/blob/main/crates/uv/src/commands/publish.rs) and [artifact validation implementation](https://github.com/astral-sh/uv/blob/main/crates/uv-publish/src/lib.rs). Verify these behaviors against the uv version selected for the follow-up.

The proposed workflow change is to trigger `publish.yml` on same-repository PRs targeting `main` with the `release` label, including updates and label changes, and remove the duplicate rehearsal from `ci.yml`. PRs would use the same detect → build → artifact download → publish path as production, with publication forced into authenticated dry-run mode. Detection must compare against the PR base SHA. Keep builds unprivileged; grant only the authentication/publish job `id-token: write` and the `pypi` environment. Actual uploads, Git tags, and GitHub releases must remain explicitly restricted to production triggers. Add a required `Publish Overall` check that accepts ordinary-PR skips and requires successful rehearsal for release PRs; `CI Overall` cannot use `needs` across workflows.

There is one repository-setting prerequisite. When checked, the `pypi` environment allowed only `main` and `elprans/fix-release-workflow`, with no required reviewers. GitHub’s normal `pull_request` workflow runs against `refs/pull/<number>/merge`: a temporary commit representing the PR merged into its target branch, without changing `main`. Adding `refs/pull/*/merge` to the environment’s deployment branch rules would permit these runs. This pattern allows all PR merge refs at the environment layer; the same-repository and `release`-label conditions belong in the workflow. See [GitHub’s environment branch rules](https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments).

The trust decision is explicit: an authenticated dry run obtains a **real, upload-capable credential**, even though uv uploads nothing. Unmerged code executing in that job could use the credential independently of our dry-run flag. The label selects eligible PRs but is not a security boundary, particularly because subsequent commits retain it. This design therefore assumes maintainer-controlled release branches are trusted to exercise production authentication. Using the same `publish.yml` and `pypi` environment matches the configured publisher identity, while the PR Git ref differs from production. See [PyPI publisher configuration](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) and [Trusted Publishing security considerations](https://docs.pypi.org/trusted-publishers/security-model/).

Finally, report the resulting signal accurately: successful token exchange confirms that PyPI recognizes the workflow’s identity and a configured publisher, but does not alone prove authorization for every standard and bundle distribution in the plan. We found no supported PyPI endpoint that gives a complete proposed-upload acceptance verdict without publishing; the documented upload API creates a release when its first file is uploaded. The follow-up should preserve the current build/install checks and add focused coverage for PR dry-run routing, authentication failures, and exclusion of upload/tag/release operations. See [PyPI token-exchange internals](https://docs.pypi.org/trusted-publishers/internals/) and [PyPI’s upload API](https://docs.pypi.org/api/upload/). No environment settings have been changed as part of this PR.

scotttrinh:
@fantix 

> We can cut the publish job into build + publish jobs in the future, so that 1) we can see which job failed, 2) built artifacts are also available on GH action run for debugging purposes, 3) we can rerun the publish job separately reusing artifacts without having to rebuild again. At that time, perhaps we don't need the publish/action.yml abstraction, given that most logic is already moved into the Python script.

Great feedback, as always. Did most of this along with some script consolidation and test pruning in https://github.com/vercel/vercel-py/pull/389/commits/71f8b317e742cfb0afaf5574d8d2bce0c16dbe07

Going to merge now, but feel free to give me any post-merge feedback and I'll incorporate it into the upcoming follow-up PR.

PR #390, opened 2026-09-12
workflow: make serialization errors fatal

fantix:
Make `SerializationError` inherit from `FatalError` so step serialization failures terminate immediately instead of retrying; retrying will just fail again so it doesn't make sense to waste that resource. It wasn't doing so before because we didn't have `FatalError` back in time. We already have the sibling-error `TypeValidationError` inheriting from `FatalError`.

This is consistent with the TS SDK.


scotttrinh (approved):
I like this pattern, nice!

msullivan (commented):
👍

PR #393, opened 2026-09-14
workflow: expose read-only hook token

fantix:
Expose `HookEvent.token` as a read-only property so callers using `BaseHook.wait()` wihtout a custom token could retrieve the generated token. Custom tokens are returned unchanged.

This matches the TypeScript hook API.


src/vercel-workflow/README.md
```diff
@@ -130,11 +130,15 @@ class Approval(BaseHook):
...
-    return approval.approved
+    async with Approval.wait() as hook:
+        print("Send approval to token:", hook.token)
+        approval = await hook
+        return approval.approved
 ```
 
 `BaseHook` supports dataclasses and Pydantic models for external resume events.
+The returned hook's read-only `token` property exposes its generated or custom
+token, which external code can pass to `resume()`.
```
scotttrinh: Do you want to add a test or example of using `resume` with a token?

scotttrinh (approved):

src/vercel-workflow/README.md
fantix: Good point! Just printing the token is too weak here.

fantix: Fixed with a bit more improvement, [preview link](https://github.com/vercel/vercel-py/blob/0b928c0895fa2e9673ffb8b9abf4d4811b2858e1/src/vercel-workflow/README.md)

PR #395, opened 2026-09-15
schedules: add vercel-schedules SDK

ricardo-agz:
Adds `vercel-schedules`, a Python port of `@vercel/schedules`, as `vercel.schedules`

```python
from datetime import datetime, timedelta
from zoneinfo import ZoneInfo
from vercel.schedules import create_schedule, list_schedules, update_schedule

schedule = await create_schedule(
    "cleanup",
    topic="scheduled-cleanup",
    cron="0 * * * *",
    timezone="America/Los_Angeles",
    jitter=timedelta(minutes=2),
    payload={"max_age_days": 30},
)

await update_schedule("cleanup", cron="0 2 * * *", jitter=None)

async for schedule in list_schedules(namespace="jobs"):
    ...
```

One-shot schedules use `at=` instead of `cron=`; they fire once and never again. A `ZoneInfo` (or UTC) datetime carries the timezone; a naive one needs `timezone=`:

```python
from datetime import datetime
from zoneinfo import ZoneInfo

# 9am Pacific on Sep 16, once
await create_schedule(
    "one-time-cleanup",
    topic="scheduled-cleanup",
    at=datetime(2026, 9, 16, 9, tzinfo=ZoneInfo("America/Los_Angeles")),
)

# same thing, naive datetime plus explicit timezone
await create_schedule(
    "one-time-report",
    topic="send-report",
    at=datetime(2026, 9, 16, 9),
    timezone="America/Los_Angeles",
)
```

On read these come back as `OneOffExpression(at=...)`, vs `CronExpression(cron=...)` for recurring schedules.

## Deliberate deviations from the TS client

- `cron=` / `at=` keywords instead of an `expression` union. `at` is a `datetime`: a `ZoneInfo` (or UTC) datetime supplies the timezone; a naive one needs `timezone=`; both together converts `at` into `timezone`. The wire `at` is the wall-clock `YYYY-MM-DDTHH:mm:ss` the service expects.
- `topic=` instead of `target: {topic}` (queue is the only creatable target). The read model keeps `QueueTarget | FunctionTarget`.
- `list_schedules` is an auto-paginating iterator; no manual cursor handling.
- `jitter` is a `timedelta` in whole minutes, 1 to 15 inclusive. VSS stores it as `jitterMinutes`, matching the AWS EventBridge Scheduler flexible time window.
- `update_schedule(jitter=None)` clears jitter; omitting it leaves it unchanged (`UNSET` sentinel).
- 404 raises `ScheduleNotFoundError`, which is also a `LookupError`.

## Inbound dispatches

`parse_schedule_event(headers, body, payload_type=...)` is framework-agnostic and validates the payload with pydantic when a type is given. `ScheduleEvent[T]` merges metadata and payload so a handler's single annotation is enough for the runtime (`resolve_payload_type`) to pick the payload type. The ASGI adapter itself lives in `vercel_runtime` (vercel/vercel-internal#945), not here.

The VSS schemas establish the remaining wire units: `createdAt` / `updatedAt` are epoch milliseconds, and one-off `at` is always a local `YYYY-MM-DDTHH:mm:ss` string without an offset. `OneOffExpression.at` is therefore naive and is interpreted in `Schedule.timezone`.


elliotdauber (commented):
Recent change we made (today) that this needs to take into account is that single one-off expressions dont take seconds, max precision is minute. Need to pass either hh:mm or hh:mm:00

src/vercel-schedules/vercel/schedules/_internal/service.py
```diff
@@ -61,8 +61,10 @@ def _timezone_name(at: datetime) -> str | None:
 
 
 def _wall_clock(at: datetime) -> str:
-    """Render a wall-clock time the way the service expects: no offset, whole seconds."""
-    return at.replace(tzinfo=None, microsecond=0).isoformat(timespec="seconds")
+    """Render a minute-aligned wall-clock time with no offset and zero seconds."""
+    if at.second or at.microsecond:
+        raise SchedulesValidationError("at must have whole-minute precision")
```
scotttrinh: Question (cc @elliotdauber): What if, instead of a runtime exception, we truncate or round the datetime?

src/vercel-schedules/vercel/schedules/_internal/models.py
```diff
@@ -0,0 +1,140 @@
...
+    """Created at runtime through this API."""
+
+
+class CronExpression(SchedulesModel):
+    """A recurring schedule, in standard five-field cron syntax.
+
+    The expression is evaluated in the schedule's `timezone`.
+    """
+
+    cron: str
```
scotttrinh: Is this where we landed? I think we had discussed something similar to APScheduler like: https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html#module-apscheduler.triggers.cron Where you have a richer, more understandable data structure with a `from_crontab(str)` initializer.

src/vercel-schedules/vercel/schedules/_internal/sentinel.py
```diff
@@ -0,0 +1,16 @@
+"""Sentinel for "argument not supplied", where `None` is itself a valid value."""
```
scotttrinh: Not a blocker: We should move this into `vercel._internal.core`. I think we have something like this in Sandbox in a bunch of places too.

PR #396, opened 2026-09-15
workflow: implement Run.terminate()

fantix:
Add `await run.terminate()` with an optional `reason`, recorded as plaintext on a `run_cancelled` event with a limit of 512 UTF-16 code units.

This is the TS SDK's `Run.cancel()`, renamed to reserve `cancel()` for cooperative cancellation. We've been missing this one until I found it during docs work.


src/vercel-workflow/vercel/workflow/_internal/world.py
```diff
@@ -622,6 +622,29 @@ def payloads(self) -> tuple[Any, ...]:
...
+class RunCancelledEventData(BaseModel):
+    cancel_reason: str | None = pydantic.Field(
+        default=None, alias="cancelReason", exclude_if=lambda e: e is None
+    )
+
+    @pydantic.field_validator("cancel_reason")
+    @classmethod
+    def validate_cancel_reason(cls, value: str | None) -> str | None:
+        # Zod's string length limit counts UTF-16 code units, like JS strings.
+        if value is not None and sum(2 if ord(char) > 0xFFFF else 1 for char in value) > 512:
```
msullivan: Maybe we want a helper for counting UTF-16 code units.
Very annoying though.

msullivan (approved):
Hm; would it be possible to come up with a modified name for this?

Maybe it would be better if we kept `cancel` meaning python task cancellation; you could imagine us making it work on workflows runs too.

Feel free to ignore me and merge if you think that's ridiculous

fantix:
> Maybe it would be better if we kept `cancel` meaning python task cancellation; you could imagine us making it work on workflows runs too.

Yeah this makes sense; since TS is already using "abort" for interrupting steps cooperatively, I would rename this to "terminate" as in "kill workflow run from outside without cleanup". We still have a chance to rename before release if you found "terminate" a bad name too.

msullivan:
> > Maybe it would be better if we kept `cancel` meaning python task cancellation; you could imagine us making it work on workflows runs too.
> 
> Yeah this makes sense; since TS is already using "abort" for interrupting steps cooperatively, I would rename this to "terminate" as in "kill workflow run from outside without cleanup". We still have a chance to rename before release if you found "terminate" a bad name too.

Terminate sounds good to me.

PR #397, opened 2026-09-15
Add sandbox Drive support

scotttrinh:
Sandbox Drives are named persistent filesystems. A Drive belongs to one project
and one region. Create or retrieve a Drive, then mount it when you create a
sandbox:

```python
from pathlib import PurePosixPath

from vercel import sandbox
from vercel.sandbox import DriveMount

cache = await sandbox.get_or_create_drive(
    name="cache",
    region="sfo1",
    max_size_bytes=10 * 1024**3,
)
async with sandbox.create_sandbox(
    region=cache.region,
    mounts={
        # Already created drive handle
        "/cache": cache,
        # Pre-existing read-only/snapshot drive by name
        "/readonly": DriveMount("shared-source", mode="snapshot"),
        # Creates a new drive by name, can use `PurePosixPath` for key
        PurePosixPath("/scratch"): "scratch",
    },
) as workspace:
    print(cache.id)
    print(workspace.mounts)

await cache.delete()
```


src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
```diff
@@ -229,6 +238,30 @@ async def delete(self) -> Self:
...
+        """Return a read-only snapshot mount for this Drive."""
+        return DriveMount(self, mode="snapshot")
+
+    async def delete(self) -> Self:
+        """Delete by project and name, then refresh this handle from the response.
+
+        Deletion does not use the Drive ID. If the name now refers to a replacement
+        Drive, this method deletes that replacement and updates this handle to it.
+        """
+        payload = await self._service.delete_drive(name=self.name, project_id=self.project_id)
```
fantix, line 260: My agent is panic here about not deleting by ID but rather by name, in case when a racing client recreated the same-name drive before our `delete()` call. Though there is no delete-by-ID API yet, shall we try the best effort to verify the ID before delete-by-name?

src/vercel-sandbox/vercel/sandbox/_internal/api_client.py
```diff
@@ -306,6 +331,11 @@ class _SandboxRoutePayload(_ApiModel):
     system: bool = False
 
 
+class _SandboxMountPayload(_ApiModel):
+    drive: str
+    mode: Literal["snapshot", "read-write"] = "read-write"
```
fantix, line 336: Should we also accept the legacy `read-only` values like the TS SDK does?

```suggestion
    mode: Literal["snapshot", "read-write"] = "read-write"

    @field_validator("mode", mode="before")
    @classmethod
    def _normalize_legacy_mode(cls, value: object) -> object:
        return "snapshot" if value == "read-only" else value
```

src/vercel-sandbox/vercel/sandbox/_internal/models.py
```diff
@@ -573,6 +575,125 @@ class DirectoryEntry:
...
+    drive_regions: set[str] = set()
+    for path, (_, value) in zip(paths, mount_entries, strict=True):
+        mount = value if isinstance(value, DriveMount) else DriveMount(value)
+        drive = mount.drive
+        name = drive if isinstance(drive, str) else drive.name
+        if name in drive_names:
+            raise ValueError(f"Drive {name!r} can be mounted only once")
+        drive_names.add(name)
+        if not isinstance(drive, str):
+            if project_id.startswith("prj_") and drive.project_id != project_id:
```
fantix, line 681: Ahh so `project_id` is actually `project_id_or_name`? What if the project name is `prj_my_proj`? It's pretty marginal, but maybe we can do this safeguard only with canonical project_id if we have one now, or delay this check after we actually have a canonical project_id (after sandbox creation, which hopefully already failed due to non-existing drive under the same name; but in the worst case, yes, the wrong drive is already mounted at the time we raise ValueError)? But anyways, I do feel like this should be resolved from the server-side.

src/vercel-sandbox/README.md
```diff
@@ -114,6 +114,74 @@ with sandbox.fork_sandbox(source_sandbox="production-agent") as forked:
...
+        PurePosixPath("/scratch"): "scratch",
+    },
+) as workspace:
+    print(cache.id)
+    print(workspace.mounts)
+
+await cache.delete()
+```
+
+Stop and destroy every sandbox that mounts a Drive before deleting it. Prefer
```
fantix: Ahhh this feels too much AI to me, with my bad English I'll do:

```suggestion
Before deleting a Drive, you must firstly stop and destroy any sandboxes that still have it mounted, like demonstrated in the `async with workspace`.
```

src/vercel-sandbox/README.md
```diff
@@ -114,6 +114,74 @@ with sandbox.fork_sandbox(source_sandbox="production-agent") as forked:
...
+`drive.delete()` when you have a handle. Deletion uses the handle's project and
+name, not its ID, so a stale handle can delete a replacement Drive with the same
+name. Use `delete_drive(name=..., project_id=...)` when creation may have
+succeeded without returning a handle.
```
fantix: This is a bit confusing - if handle deletion may do the wrong thing, why is it still preferred? And I got lost reading "creation may have succeeded without returning a handle" too.

Then I prompted my agent with like "I would reword this sentence like this xxxx to be more natural, so that readers without implicit context would understand; list other places where you would leave a similar comment in the PR" (my actual prompt is in Chinese so this is roughly it), but I didn't finish reading the full feedback or copy over ;)

fantix (approved):
No blocking issue for me :lfg:

src/vercel-sandbox/vercel/sandbox/_internal/async_runtime.py
scotttrinh: Mine did too lmao! I _think_ the name is meant to be the identifier similar to how sandboxes are identified by a name, such that if you did replace a drive and then called delete from a separate concurrent request, it would delete the new drive rather than the one you pulled. This behavior is consistent with the TS SDK, and seems to be working as intended, but I'll bring this up with the team and see if we want to rethink how this ought to work.

src/vercel-sandbox/vercel/sandbox/_internal/api_client.py
scotttrinh: I considered that but I don't think we should since it's not a backwards-compatibility issue.

src/vercel-sandbox/vercel/sandbox/_internal/models.py
scotttrinh: Yeah, this also felt kinda brittle to me, but is consistent with how we do this kind of detection client-side. We can beef this up later.

src/vercel-sandbox/README.md
fantix: OK now I understand, deleting a replacement Drive is intentional, not a mistake. But the last sentence still feels irrelevant.

scotttrinh: Gotcha. Lemme give it more of a human touch here and see if I can make this more understandable, thanks for pointing this out!

